PR #1, opened 2018-05-10
Update for Elm 0.19

eeue56:


src/Expect.elm
```diff
@@ -362,12 +362,12 @@ within tolerance a b =
 {-| Passes if (and only if) a call to `within` with the same arguments would have failed.
 -}
 notWithin : FloatingPointTolerance -> Float -> Float -> Expectation
-notWithin tolerance a b =
```
mgold: Let's change this for `within` as well.

src/Expect.elm
```diff
@@ -394,6 +394,7 @@ true : String -> Bool -> Expectation
 true message bool =
     if bool then
         pass
+
```
mgold: I'm assuming these are added by elm-format.

src/Fuzz.elm
```diff
@@ -45,8 +46,8 @@ import Fuzz.Internal as Internal
         , invalidReason
         )
 import Lazy
-import Lazy.List exposing ((+++), LazyList)
-import Random.Pcg as Random exposing (Generator)
+import Lazy.List exposing (LazyList, append)
+import Random exposing (Generator)
```
mgold: Let's also update the docs around line 65 of this file to refer to the new library.

src/Test.elm
```diff
@@ -389,7 +395,7 @@ fuzz2 fuzzA fuzzB desc =
         fuzzer =
             Fuzz.tuple ( fuzzA, fuzzB )
     in
-    uncurry >> fuzz fuzzer desc
+    (\f ( a, b ) -> f a b) >> fuzz fuzzer desc
```
mgold: `(\f (a,b) -> fuzz fuzzer desc (f a b))`?

src/Test.elm
```diff
@@ -412,63 +418,10 @@ fuzz3 fuzzA fuzzB fuzzC desc =
     uncurry3 >> fuzz fuzzer desc
 
 
-{-| Run a [fuzz test](#fuzz) using four random inputs.
-
-This is a convenience function that lets you skip calling [`Fuzz.tuple4`](Fuzz#tuple4).
-
--}
-fuzz4 :
-    Fuzzer a
```
mgold: This is the first I've heard of this API change. I'm okay with it, but I'm not sure it should be bunched into the 0.19 upgrade.

We have some other major changes we want to make, and we should also have a version where these functions are deprecated.

avh4: Tuples of size 4 and larger are disallowed by elm-compiler in 0.19, so uncurry4/5 and tuple4/5 have to be removed.  (Though I suppose fuzz4/5 could technically stay, but would be slightly messy to reimplement without tuple/uncurry.)

mgold: On second thought, I'm okay with one big release. Let's just be sure to look at the *major release blocker* keyword on the old repo

mgold: Makes sense. fuzz4/5 are also not great since shrinking is hard.

src/Fuzz.elm
mgold: 3b070c1

src/Expect.elm
mgold: 3b070c1

mgold:
I think we're good for an alpha release to get other packages unblocked.

eeue56:
Not ready for merge/publishing

avh4:
#2 adds on to this branch

avh4:
From slack:

> eeue56 [12:23 PM]
> I think we need to merge elm-html-test at some point, but it probably doesnt need to be before 1.0.0
> 
> (we can merge it before the non-alpha)
> But actually now the majority of stuff to do is the CLI, which Richard is looking at

PR #2, opened 2018-05-18
Tests compile with 0.19

avh4:


tests/FuzzerTests.elm
```diff
@@ -15,13 +16,6 @@ die =
     Fuzz.intRange 1 6
 
 
-seed : Fuzzer Random.Seed
-seed =
```
avh4: `Helpers.randomSeedFuzzer` was used on all but one test in here; I just replaced the one use of `seed` with `randomSeedFuzzer`, but maybe there was a reason it needed something different?

tests/FuzzerTests.elm
```diff
@@ -170,16 +142,14 @@ shrinkingTests =
...
                                     else
                                         checkPair (b :: more)
 
                                 _ ->
                                     True
                     in
                     checkPair aList |> Expect.true "[1,0]|[0,-1]"
-            , fuzz (intRange 1 8 |> andThen (\i -> intRange 0 (2 ^ i))) "Fuzz.andThen shrinks a number" <|
-                \i ->
-                    i <= 2 |> Expect.true "3"
```
avh4: `Fuzz.andThen` seems to be gone in master.

tests/Runner/Log.elm
```diff
@@ -71,12 +72,13 @@ logOutput summary =
         _ =
             if summary.failed > 0 || summary.autoFail /= Nothing then
                 output
-                    |> flip Debug.log 1
-                    |> (\_ -> Debug.crash "FAILED TEST RUN")
+                    |> (\a -> Debug.log a 1)
+                    |> (\_ -> Debug.todo "FAILED TEST RUN")
```
avh4: I didn't understand when this is meant to occur; probably something better can/should be done here.

avh4: Ah, this seems to be how the elm-test test suite causes Platform.worker to exit and to give a non-zero exit code.   Does the elm-test CLI depend on this for the same reason?  Seems like eventually this will need to be replaced by ports, but maybe using `Debug.todo` is good enough for now?

mgold: The exit code indicates (to CI systems) whether the tests passed or failed.

tests/FuzzerTests.elm
mgold: I think this can be removed safely.

tests/Runner/Log.elm
avh4: 👍  then I think `Debug.todo` is fine for this PR, but might need to revisit this in #3 or once we get to hooking up the CLI again.

PR #3, opened 2018-05-18
Tests pass in Elm 0.19 (and implements `toString`)

avh4:
~Depends on #2; this PR starts at 18444a9~ edit: rebased onto `elm-explorations:0.19` after #2 got merged.

~This seems to be blocked by Platform.worker triggering a compiler bug: https://github.com/elm-lang/elm-compiler/issues/1706~

~It appears that `elm make` only allows package projects to contain kernel code (since application projects don't have a name, so it doesn't know we're an elm-explorations project); so this tries to convert the tests to a package project, but that means we can't have multiple source-directories, so a bunch of symlinks into ../src/* are added.  With `Main.main = Platform.worker` commented out (see https://github.com/elm-lang/elm-compiler/issues/1706), everything compiles, but `elm make` doesn't write an output file if there's no `main`!~

  - [x] change `Main.main` to be a Browser.program (or wait for https://github.com/elm-lang/elm-compiler/issues/1706 to be fixed), then we should be able to get an output js file, and try to run it in node or have a js file that calls something in the compiled elm js to evaluate the tests.
  - [x] uncomment the SeedTests (in Main.elm), and
     - [x] figure out why it doesn't print the name of the failing tests
     - [x] figure out why some tests are failing
  - ~make sure tests pass on Travis https://travis-ci.org/avh4/new-elm-test/branches~ edit: this isn't easily possible because the elm binaries aren't easily installable

Since CI isn't usable right now, tests can be verified as follows:
```sh
# !!! this will delete all dirty and untracked files !!!
git clean -dfx
nvm use 10  # optional, or repeat with "4" to mimic travis
cd tests
npm install
npm test
cd ..
```


tests/src/Diff.elm
```diff
@@ -12,4 +12,4 @@ type Change a
 
 diff : List a -> List a -> List (Change a)
 diff _ _ =
-    Debug.todo "Upgrade jinjor/elm-diff for Elm 0.19"
+    []
```
avh4: TODO

tests/src/Runner/Log.elm
```diff
@@ -73,7 +73,7 @@ logOutput summary =
             if summary.failed > 0 || summary.autoFail /= Nothing then
                 output
                     |> (\a -> Debug.log a 1)
-                    |> (\_ -> Debug.todo "FAILED TEST RUN")
+                    -- |> (\_ -> Debug.todo "FAILED TEST RUN")
```
avh4: TODO: probably put this back

tests/src/Tests.elm
```diff
@@ -18,26 +18,28 @@ all : Test
...
-        , fuzzerTests
-        , floatWithinTests
-        , RunnerTests.all
+
+        -- , regressions
+        -- , testTests
+        -- , expectationTests
+        -- , fuzzerTests
+        -- , floatWithinTests
+        -- , RunnerTests.all
```
avh4: TODO: revert this

tests/src/Tests.elm
```diff
@@ -18,26 +18,28 @@ all : Test
...
     describe "The String module"
         [ describe "String.reverse"
-            [ test "has no effect on a palindrome" <|
-                \_ ->
-                    let
-                        palindrome =
-                            "hannah"
-                    in
-                    Expect.equal palindrome (String.reverse palindrome)
+            [ only <|
```
avh4: TODO: remove `only`

src/Elm/Kernel/TestInternal.js
```diff
@@ -0,0 +1,9 @@
+/*
+
+import Elm.Kernel.Debug exposing (toString)
+
+*/
+
+function _TestInternal_toString(value) {
+  __Debug_toString(value);
+}
```
eeue56: Do we need this? I don't think we do -- I think we can just import `Elm.Kernel.Debug` in the elm code

avh4: I thought about that but didn't try it yet;  I assumed that it wouldn't work because Elm.Kernel.Debug isn't an exposed module of core, but I will try it.

eeue56: The way it used to work, at least, was that any native code could be imported anywhere if you had permission to do so (via native-modules). I imagine it should be the same for kernel code -- but I could be wrong.

avh4: Just tried, and yes it seems to work the way you say!

tests/src/SeedTests.elm
```diff
@@ -94,19 +95,19 @@ tests =
...
             \num ->
-                Expect.equal num 409469537
+                Expect.equal num -8
         , describe "Seed test"
             [ fuzzTest ]
         , describe "another top-level fuzz test"
             [ fuzz int "it still gets different values, due to computing the seed as a hash of the label, and these labels must be unique" <|
                 \num ->
-                    Expect.equal num 0
+                    Expect.equal num -1079557009
```
avh4: It seems suspicious that `0` was the old value here; was that actually the first value fuzz was generating for this test, or was the zero significant for some other reason?  I didn't really understand what the test is saying it's meant to check.

tests/src/SeedTests.elm
```diff
@@ -1,22 +1,23 @@
...
 
 
+
 -- NOTE: These tests are only here so that we can watch out for regressions. All constants in this file are what the implementation happened to output, not what we expected the implementation to output.
 
 
 expectedNum : Int
 expectedNum =
-    -3954212174
+    20
```
avh4: The values generated by a particular seed are different now, after the switch from `Random.Pcg` to `Random`.  That's to be expected, right?

tests/src/SeedTests.elm
mgold: I know it has to do with [generating the seeds based on a hash of the descriptions](https://github.com/elm-community/elm-test/issues/238#issuecomment-386096275). Whether zero is important or not, I don't know.

tests/src/SeedTests.elm
mgold: That's slightly suspicious. I'm not sure though.

mgold: I've confirmed that, for identical initial seeds an identical generator, PCG in 0.18 and elm-lang/random in 0.19 produce identical output. So we may have a bug in elm-test where something is getting changed.

eeue56: Could be in the shrinker. I don't think of the logic changed, however.

avh4: Hmm, in this branch, the test is getting `generate (Seed 1749620150 3558552409) int = 20`.  In 0.18, elm-community/elm-test#master, `generate (Seed 3075227300 1013904223) int = -3954212174`.  It seems like something changed in distributing the seeds...

avh4: the change seems to be due to https://github.com/elm-lang/random/commit/8796daa0b7a7d2c00bab6a12300ec838e20c09fb

avh4: Okay, I was able to fix this with https://github.com/elm-explorations/test/pull/3/files#diff-ef115dbbb74efc849f44b2da296b93a9R244

src/Test/Runner.elm
```diff
@@ -241,6 +241,13 @@ distributeSeedsHelp hashed runs seed test =
             let
                 ( firstSeed, nextSeed ) =
                     Random.step Random.independentSeed seed
+                        |> maintainElm18RandomPcgBehavior
+
+                -- In https://github.com/elm-lang/random/commit/8796daa0b7a7d2c00bab6a12300ec838e20c09fb,
+                -- the output of Random.independentSeed was changed.  This function is applied to the result
+                -- so that seed values given by old version of elm-test still give the same test results.
+                maintainElm18RandomPcgBehavior ( a, b ) =
+                    ( b, a )
```
avh4: I think this fix only is one level deep:  the values generated from the independent seed will now match what the would be in 0.18 PCG;  **but** the "incr" of the nextSeed will be different.  I think that might mean that subsequent independent seeds made from the nextSeed will be different?  The current test suite doesn't seem to exercise that case.  To make it exactly the same, we need to advance the nextSeed by one step, but the public API of Random doesn't provide a way to do this (Random.float advances by two steps, and Random.int advances by a random number of steps).

drathier: I'd vote for 1 or 3. Making the tests look like they're the same most of the time means people will think that they are always the same.

rtfeldman: I think it's fine if we don't apply this fix.

It doesn't seem reasonable to me to expect test failures from 0.18 `elm-test` to be reproducible using the same `--seed` in 0.19 `elm-test`.

Even if it were, the way that feature is typically used is in conjunction with a particular code commit, e.g. "to reproduce, run this git sha with this `--seed` value" - and in that case the git sha would of course be on 0.18 if the `--seed` were from 0.18 `elm-test`.

I think we should let it be different!

tests/src/SeedTests.elm
avh4: We decided to remove this fix, see https://github.com/elm-explorations/test/pull/3#discussion_r189445609

PR #4, opened 2018-05-22
Fix major release blockers

mgold:
As asked on Slack, here are fixes for issues tagged major-release-blocker in the main repo.

There is also mention in 191 (no link for stealth) of removing `Fuzz.conditional`. This is a pretty clunky filtering system that I don't think we should encourage. Anyone feel strongly in favor of keeping it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 | Version                                                              | Notes                                                                                                                                            |
 | -------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
-| [**5.0.0**](https://github.com/elm-explorations/elm-test/tree/5.0.0) | Update for Elm 0.19. Remove `Fuzz.andThen` and `Test.Runner.getFailure`.                                                                         |
+| [**5.0.0**](https://github.com/elm-explorations/elm-test/tree/5.0.0) | Update for Elm 0.19. Remove `Fuzz.andThen` and `Test.Runner.getFailure`. Fail on equating floats to encourage checks with tolerance.             |
 | [**4.0.0**](https://github.com/elm-explorations/elm-test/tree/4.0.0) | Add `only`, `skip`, `todo`; change `Fuzz.frequency` to fail rather than crash on bad input, disallow tests with blank or duplicate descriptions. |
 | [**3.1.0**](https://github.com/elm-explorations/elm-test/tree/3.1.0) | Add `Expect.all`                                                                                                                                 |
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
...
-| [**3.0.0**](https://github.com/elm-explorations/elm-test/tree/3.0.0) | Update for Elm 0.18; switch the argument order of `Fuzz.andMap`.
-| [**2.1.0**](https://github.com/elm-explorations/elm-test/tree/2.1.0) | Switch to rose trees for `Fuzz.andThen`, other API additions.
-| [**2.0.0**](https://github.com/elm-explorations/elm-test/tree/2.0.0) | Scratch-rewrite to project-fuzzball
-| [**1.0.0**](https://github.com/elm-explorations/elm-test/tree/1.0.0) | ElmTest initial release
+
+| Version                                                              | Notes                                                                                                                                            |
+| -------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| [**5.0.0**](https://github.com/elm-explorations/elm-test/tree/5.0.0) | Update for Elm 0.19. Remove `Fuzz.andThen` and `Test.Runner.getFailure`. Fail on equating floats to encourage checks with tolerance.             |
```
avh4: This will actually be `elm-explorations/test 1.0.0`.  How do we want to format the changelog for the package rename?

CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
...
+| [**4.0.0**](https://github.com/elm-explorations/elm-test/tree/4.0.0) | Add `only`, `skip`, `todo`; change `Fuzz.frequency` to fail rather than crash on bad input, disallow tests with blank or duplicate descriptions. |
```
avh4: Tags 4.0.0 and earlier should link to elm-community/elm-test instead of elm-explorations.

CHANGELOG.md
mgold: I'm not sure, and it's probably beyond the scope of this PR.

avh4: I just added https://github.com/elm-explorations/test/pull/4/commits/e36519d68e3830e8c52efd564878b2544f8a02dd

CHANGELOG.md
mgold: I'll fix this, but the changelog was preexisting. You may have missed that because my autoformatter redid the whole file.

avh4:
gonna merge this so I can continue with getting this published as alpha to unblock the work on node-test-runner.  I think we can still make major changes during the alpha before the final 1.0.0 release if we want to.

PR #6, opened 2018-05-22
Remove Debug.todo (option 1: Test.Runner.fuzz returns Result)

avh4:
(required by Elm 0.19 in order to publish packages)

### DIFF
--- a/src/Test/Runner.elm
+++ b/src/Test/Runner.elm
@@ -465,7 +465,7 @@ type Shrinkable a
 {-| Given a fuzzer, return a random generator to produce a value and a
 Shrinkable. The value is what a fuzz test would have received as input.
 -}
-fuzz : Fuzzer a -> Random.Generator ( a, Shrinkable a )
+fuzz : Fuzzer a -> Result String (Random.Generator ( a, Shrinkable a ))
 fuzz fuzzer =
     case fuzzer of
         Ok validFuzzer ->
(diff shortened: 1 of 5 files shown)


src/Test/Runner.elm
```diff
@@ -465,7 +465,7 @@ type Shrinkable a
 {-| Given a fuzzer, return a random generator to produce a value and a
 Shrinkable. The value is what a fuzz test would have received as input.
 -}
-fuzz : Fuzzer a -> Random.Generator ( a, Shrinkable a )
+fuzz : Fuzzer a -> Result String (Random.Generator ( a, Shrinkable a ))
```
avh4: Note: this is a major API change.

I think ideally, instead functions that can create invalid fuzzers should instead return Result, and this function wouldn't need to.  I didn't look into the details yet of how invalid fuzzers can be created to see if that even makes sense.

avh4: See https://github.com/elm-explorations/test/pull/7 for what I think is probably a better approach.

mgold: This function was added because ... I think someone requested it. Point being it's an expert thing and I learn towards keeping invalid fuzzers at the cost of making the expert/rare API trickier.

rtfeldman: I don't think a breaking API change to runners is a big deal, and I remember the concept of invalid fuzzers being worth adding.

I think we should do it this way!

avh4:
TODO: update the changelog

src/Test/Runner.elm
avh4: sounds good!  I also originally thought that the official test runners depended on this, but it appears they don't (not that that would have been a big issue anyway), so I'll go with Max/Richard's conclusion.

PR #7, opened 2018-05-22
Remove Debug.todo (option 2: remove invalid fuzzers)

avh4:
This is an alternative to #6.  In this case, the concept of an "invalid fuzzer" is removed.


src/Fuzz.elm
```diff
@@ -561,60 +548,30 @@ so:
...
-        aList
-            |> List.map extractValid
-            |> combineValid
-            |> Result.map Random.frequency
-
-
-extractValid : ( a, Valid b ) -> Valid ( a, b )
-extractValid ( a, valid ) =
-    Result.map (\b -> ( a, b )) valid
+frequency : ( Float, Fuzzer a ) -> List ( Float, Fuzzer a ) -> Fuzzer a
```
avh4: This is the major API change here; another option here is to have `frequency` return `Result`, but that seemed less nice to me.

mgold: We had a variation of this and it was painful to work with. I know Richard is out but I want to hear his opinion on this.

avh4:
Closed in favor of #6:  https://github.com/elm-explorations/test/pull/6/files#r189778486

PR #9, opened 2018-05-24
Move native code from node-test-runner

avh4:
0.19 doesn't seem to like compiling something with both local kernel code and a `main` program.  So to make this work, I had to set up the tests in a way that means they can no longer use functions that aren't publicly exposed.  Luckily there were only a handful of tests that were affected by that, and I think it's probably better in the end to have the test suite only exercise the public API anyway.


drathier:
Ouch. Do we have any way to see if a test isn't exposed, so we can fail a test run as incomplete until everything is exposed?

avh4:
> Do we have any way to see if a test isn't exposed

I'm not sure I understood what you meant.   You're saying, is there a way to mark the tests as not working other than commenting them out?  I was planning to finish fixing all of them before considering merging this to master.

src/Elm/Kernel/Test.js
```diff
@@ -0,0 +1,20 @@
...
+
+function _Test_runThunk(thunk)
+{
+  try {
+    // Attempt to run the thunk as normal.
+    return thunk(__Utils_Tuple0);
+  } catch (err) {
+    // If it throws, return a test failure instead of crashing.
+    var message = 'This test failed because it threw an exception: "' + err + '"';
+    return __List_Cons(__Expect_fail(message), __List_Nil);
```
eeue56: I wonder if it's better just to return the error message, removing the need for importing Expect in here at all.

drathier:
I understood it as: if there's a test file with a test that isn't exposed, we cannot run it. So forgetting to expose a test would mean it would never run, which is way worse than commenting out test cases. Thus, I wondered if we can at least know that there's something in there that isn't exposed, so we can warn users about this?

src/Elm/Kernel/Test.js
avh4: so then it would be `runThunk : (() -> a) -> Result String a` ?  I can give that a try.

eeue56: Yep!

avh4: Changed in 6568242

avh4:
@drathier I think your concern is something that was still a problem previously; this PR doesn't do anything to change that.

In any case, what you're asking about would only affect the test suite of this repo; normal users of elm-test will use the CLI, which will automatically include all exposed tests, so there's nothing to worry about there.

avh4:
I'm going to publish this so I can continue working on node-test-runner.  If you have concerns with anything here, comment here, or open a new issue and @-mention me.

PR #21, opened 2018-09-04
Fix formatting of doc comment for Fuzz.andThen

harrysarson:
Adds back ticks and indents the example to ensure correct rendering.

Hopefully, this will make the doc a bit easier to read 😄


mgold:
Failure seems to be because: `{"path":"src/Fuzz.elm","message":"File is not formatted with elm-format-0.8.0 --elm-version=0.19"}`

harrysarson:
Ah 😨 A downside of using the github editor. Will clone and fix.

mgold:
Thanks!

PR #35, opened 2018-09-28
attempting to explain the shrinkers better

Bastes:
This is more a first draft than anything else, but it should be enough to start discussing :)

Don't hesitate to comment, criticize and discuss.

### DIFF
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -382,9 +382,10 @@ Takes a tuple of shrinkers and returns a shrinker of tuples.
 -}
 tuple : ( Shrinker a, Shrinker b ) -> Shrinker ( a, b )
 tuple ( shrinkA, shrinkB ) ( a, b ) =
-    Lazy.List.map2 Tuple.pair (shrinkA a) (shrinkB b)
-        |> append (Lazy.List.map (\a2 -> ( a2, b )) (shrinkA a))
-        |> append (Lazy.List.map (\b2 -> ( a, b2 )) (shrinkB b))
+    append (Lazy.List.map (Tuple.pair a) (shrinkB b))
+        (append (Lazy.List.map (\first -> ( first, b )) (shrinkA a))
+            (Lazy.List.map2 Tuple.pair (shrinkA a) (shrinkB b))
+        )
 
 
 {-| 3-Tuple shrinker constructor.
@@ -392,13 +393,18 @@ Takes a tuple of shrinkers and returns a shrinker of tuples.
 -}
 tuple3 : ( Shrinker a, Shrinker b, Shrinker c ) -> Shrinker ( a, b, c )
 tuple3 ( shrinkA, shrinkB, shrinkC ) ( a, b, c ) =
-    Lazy.List.map3 (\a2 b2 c2 -> ( a2, b2, c2 )) (shrinkA a) (shrinkB b) (shrinkC c)
-        |> append (Lazy.List.map2 (\a2 b2 -> ( a2, b2, c )) (shrinkA a) (shrinkB b))
-        |> append (Lazy.List.map2 (\a2 c2 -> ( a2, b, c2 )) (shrinkA a) (shrinkC c))
-        |> append (Lazy.List.map2 (\b2 c2 -> ( a, b2, c2 )) (shrinkB b) (shrinkC c))
-        |> append (Lazy.List.map (\a2 -> ( a2, b, c )) (shrinkA a))
-        |> append (Lazy.List.map (\b2 -> ( a, b2, c )) (shrinkB b))
-        |> append (Lazy.List.map (\c1 -> ( a, b, c1 )) (shrinkC c))
+    append (Lazy.List.map (\c1 -> ( a, b, c1 )) (shrinkC c))
+        (append (Lazy.List.map (\b2 -> ( a, b2, c )) (shrinkB b))
+            (append (Lazy.List.map (\a2 -> ( a2, b, c )) (shrinkA a))
+                (append (Lazy.List.map2 (\b2 c2 -> ( a, b2, c2 )) (shrinkB b) (shrinkC c))
+                    (append (Lazy.List.map2 (\a2 c2 -> ( a2, b, c2 )) (shrinkA a) (shrinkC c))
+                        (append (Lazy.List.map2 (\a2 b2 -> ( a2, b2, c )) (shrinkA a) (shrinkB b))
+                            (Lazy.List.map3 (\a2 b2 c2 -> ( a2, b2, c2 )) (shrinkA a) (shrinkB b) (shrinkC c))
+                        )
+                    )
+                )
+            )
+        )
 
 
 
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -57,48 +57,52 @@ Let's say I'm writing a Shrinker for binary trees :
         = Node (Tree a) (Tree a)
         | Leaf a
 
-In this context, "smaller" should probably mean "with fewer nodes", to make the
-reason the test failed more obvious:
+Now let's say the test fails with the following randomly-generated tree:
 
-    -- shrinking from this failing input:
     Node
-    ( Node
-      ( Node
-        ( Node
-          ( Leaf 888 )
-          ( Leaf 9090 )
+        (Node
+            (Node
+                (Node
+                    (Leaf 888)
+                    (Leaf 9090)
+                )
+                (Node
+                    (Leaf -1)
+                    (Node
+                        (Leaf 731)
+                        (Node
+                            (Leaf 9621)
+                            (Leaf -12)
+                        )
+                    )
+                )
+            )
+            (Node
+                (Leaf -350)
+                (Leaf 124)
+            )
         )
         (Node
-          ( Leaf -1 )
-          ( Node
-            ( Leaf 731 )
-            ( Node
-              ( Leaf 9621 )
-              ( Leaf -12 )
+            (Leaf 45)
+            (Node
+                (Leaf 123)
+                (Node
+                    (Leaf 999111)
+                    (Leaf -148148)
+                )
             )
-          )
         )
-      )
-      ( Node
-        ( Leaf -350 )
-        ( Leaf 124 )
-      )
-    )
-    ( Node
-      ( Leaf 45 )
-      ( Node
-        ( Leaf 123 )
-        ( Node
-          ( Leaf 999111 )
-          ( Leaf -148148 )
-        )
-      )
-    )
-    -- to this failing input:
+
+This is a pretty big tree, there are many nodes and leaves, and it's difficult
+to tell which is responsible for the failing.
+
+Now let's say we shrink it to the following simpler value:
+
     Leaf -1
-    -- it will probably take less time for the developer inspecting the failing
-    -- input to understand what went wrong (probably when there's a `Leaf -1` in
-    -- the tree)
+
+It will probably take less time for the developer inspecting the failing input
+to understand what went wrong (probably when there's a `Leaf` holding a negative
+number in the tree).
 
 
 ### How does shrinking work?
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -120,10 +120,10 @@ failing value is presented instead.
 
 ### How do I make my own Shrinkers?
 
-Shrinkers must be deterministic, since they do not have access to a random
-number generator. It's the generator part of the fuzzer that's meant to find
-rare edge cases; it's the shrinkers job to make the failures as understandable
-as possible.
+Shrinkers are deterministic, since they do not have access to a random number
+generator. It's the generator part of the fuzzer that's meant to find the rare
+edge cases; it's the shrinkers job to make the failures as understandable as
+possible.
 
 Shrinkers have to return a LazyList, something that works a bit like a list.
 That LazyList may or may not have another element each time we ask for one,
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -57,7 +57,7 @@ Let's say I'm writing a Shrinker for binary trees :
         = Node (Tree a) (Tree a)
         | Leaf a
 
-In this context, "smaller" should probably mean "with less nodes", to make the
+In this context, "smaller" should probably mean "with fewer nodes", to make the
 reason the test failed more obvious:
 
     -- shrinking from this failing input:
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -104,7 +104,7 @@ reason the test failed more obvious:
 ### How does shrinking work?
 
 When a test fails with a random value, this value is fed to the shrinker which
-returns a LazyList of potentially failing inputs. Those will be passed to the
+returns a bunch of potentially failing inputs. Those will be passed to the
 test that failed given that original failing input.
 
 The fuzz test is rerun with the first "smaller" value. If the test now passes,
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -95,7 +95,7 @@ of exceptions.
     -- make a test loop indefinitely:
     shrinker 123 == [ 456, ... ]
     shrinker 456 == [ 789, ... ]
-    shrinker 789 == [ 123, ... ] -- here we re-produce our first value and loop
+    shrinker 789 == [ 123, ... ] -- here we reproduce our first value and loop
 
 
 # Shrinking Basics
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -61,7 +61,7 @@ It's a function that takes an original failing input and gives back a LazyList
 of potentially failing inputs. Those will be passed to the test that failed
 given that original failing input.
 
-The fuzzy test is rerun with the first "smaller" value. If the test now passes,
+The fuzz test is rerun with the first "smaller" value. If the test now passes,
 we discard that smaller value and try the next one in the list. If instead the
 test still fails, the rest of the list is discarded and that "smallest" value is
 recursively shrunk.
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -58,7 +58,7 @@ That's kind of arbitrary, and depends on what kind of values you're fuzzing.
 When you write your own Shrinker, you decide what is small for the kind of data
 you're testing with.
 
-Let's say I'm writing a Shrinker for binary trees :
+Let's say I'm writing a Shrinker for binary trees:
 
     -- randomly-generated binary trees might soon become unreadable
     type Tree a
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -88,7 +88,7 @@ at the shrinking phase).
 
 Also, the shrinking process should never produce the same value twice, even
 through several shrinking attempts. Doing so may result in tests looping over
-wrong values that return endlessly, hanging or generating stack too deep kind
+wrong values, hanging or generating stack too deep kind
 of exceptions.
 
     -- so, for example, a shrinker producing these outputs from these inputs could
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -61,7 +61,39 @@ In this context, "smaller" should probably mean "with fewer nodes", to make the
 reason the test failed more obvious:
 
     -- shrinking from this failing input:
-    Node (Node (Node (Leaf 8) (Node (Leaf -1) (Leaf 7))) (Node (Leaf 3))) (Leaf 4)
+    Node
+    ( Node
+      ( Node
+        ( Node
+          ( Leaf 888 )
+          ( Leaf 9090 )
+        )
+        (Node
+          ( Leaf -1 )
+          ( Node
+            ( Leaf 731 )
+            ( Node
+              ( Leaf 9621 )
+              ( Leaf -12 )
+            )
+          )
+        )
+      )
+      ( Node
+        ( Leaf -350 )
+        ( Leaf 124 )
+      )
+    )
+    ( Node
+      ( Leaf 45 )
+      ( Node
+        ( Leaf 123 )
+        ( Node
+          ( Leaf 999111 )
+          ( Leaf -148148 )
+        )
+      )
+    )
     -- to this failing input:
     Leaf -1
     -- it will probably take less time for the developer inspecting the failing
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -70,7 +70,7 @@ Once we can't find anymore "smaller" shrunken values that make the test fail,
 the smallest we found thus far is presented as the smallest failing value.
 
 
-## How to make you own shrinkers?
+## How do I make my own Shrinkers?
 
 Shrinkers should be deterministic, so no random generation here.
 
--- a/src/Shrink.elm
+++ b/src/Shrink.elm
@@ -109,7 +109,7 @@ Now let's say we shrink it to the following simpler value:
     Leaf -1
 
 It will probably take less time for the developer inspecting the failing input
-to understand what went wrong. Next, the developer could check wether a negative
+to understand what went wrong. Next, the developer could check whether a negative
 number in a leaf is what causes the test to fail.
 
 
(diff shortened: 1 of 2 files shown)


src/Shrink.elm
```diff
@@ -8,6 +8,62 @@ module Shrink exposing
...
+
+## How do I make my own shrinkers?
+
+Shrinkers should be deterministic, so no random generation here.
+
+Shrinkers have to return a LazyList, something that works a bit like a list
+that may or may not have another element each time we ask for one, but doesn't
+necessarily have them all committed to memory, occupying less space (especially
+interesting since there may quite a lot of elements, maybe and infinite number
+of them).
```
mgold: "Maybe an" not "Maybe and" but more importantly, the shrinker should not propose an infinite number of possible shrinks. It should propose a *constant* number of shrinks.

There's two schools of thought here: try to shrink a value aggressively, or try to shrink it incrementally and rely on the recursive nature of the shrinking algorithm. (That's a good distinction: the shrinking algorithm is generic and implemented in the library, to use the shrinker of a type effectively.) An aggressive list shrinker would try the empty list and the singleton of the head before anything else. An incremental shrinker might try the empty list only if given a singleton, and otherwise try the singleton of the head and the tail. If the empty list causes a test failure, both approaches will get there. I don't have hard evidence on which to prefer, but I intuit incremental is better. I'm sure if this whole distinction is worth documenting or not.

Also, this is another run-on.

src/Shrink.elm
```diff
@@ -33,18 +89,21 @@ import String
...
 
 {-| Perform shrinking. Takes a predicate that returns `True` if you want
-shrinking to continue (e.g. the test failed). Also takes a shrinker and a value
-to shrink. It returns the shrunken value, or the input value if no shrunken
+shrinking to continue (most likely the failing test for which we are attempting
+to shrink the value). Also takes the shrinker and the value to shrink.
+
+It returns the shrunken value, or the input value if there are no shrunken
 values that satisfy the predicate are found.
+
```
mgold: I wonder if this function should me moved to an internal fuzz module in the next major release. It's not really part of writing shrinkers; it's how shrinkers get used.

src/Shrink.elm
```diff
@@ -8,6 +8,62 @@ module Shrink exposing
...
+## Soooo a Shrinker is?...
+
+It's a function that takes an original failing input and gives back a LazyList
+of potentially failing inputs that will be tested against through the test that
+has failed given that original failing input.
+
+Once all of these potentially failing inputs have been tested, the "smallest"
+one that still fails will be presented to the tester for analysis of what went
+wrong.
```
mgold: Not all potentially failing inputs. How about: The fuzz test is rerun with the first "smaller" value. If the test now passes, then we discard that smaller value and try the next one in the list. If instead the test still fails, the rest of the list is discarded and that value is recursively shrunk.

src/Shrink.elm
Bastes: I'm not sure what you meant about the two schools of thoughts ^^° so I'm reworking this paragraph and leaving that distinction for later when I get what it is (and wether it's something that helps understand shrinking better or something that can be learned later on by someone who really means to dive into it).

:)

src/Shrink.elm
Bastes: I'm on two minds on this ; on on hand it helps understanding what shrinkers are used for, ont the other it's indeed not necessary to expose for one to make one's custom shrinkers ; either way, the bit of documentation can't hurt :)

src/Shrink.elm
Bastes: Sound great :) thanks.

Bastes:
Here's round 2: less informal (and hilarious ;p) and taking into account most if not all of your comments (especially about run-on sentences ;p).

harrysarson:
Maybe it is worth adding a section: `Why does elm test need shrinkers?`.

I think the answer to the question is already covered but it might be nice to move it to an explicit section.

Bastes:
Here's round 3: with an explanation on why shrinkers are needed :)

src/Shrink.elm
```diff
@@ -8,6 +8,86 @@ module Shrink exposing
...
+    -- input to see what went wrong (probably the second leaf of the root node)
+
+
+## What is a Shrinker?
+
+It's a function that takes an original failing input and gives back a LazyList
+of potentially failing inputs. Those will be passed to the test that failed
+given that original failing input.
+
+The fuzzy test is rerun with the first "smaller" value. If the test now passes,
```
mgold: fuzzy -> fuzz

src/Shrink.elm
```diff
@@ -8,6 +8,86 @@ module Shrink exposing
...
+we discard that smaller value and try the next one in the list. If instead the
+test still fails, the rest of the list is discarded and that "smallest" value is
+recursively shrunk.
+
+Once we can't find anymore "smaller" shrunken values that make the test fail,
+the smallest we found thus far is presented as the smallest failing value.
+
+
+## How to make you own shrinkers?
```
mgold: "How do I make my own Shrinkers" maybe? I'm not sure what we're capitalizing in titles.

src/Shrink.elm
```diff
@@ -8,6 +8,86 @@ module Shrink exposing
...
+
+Shrinkers should be deterministic, so no random generation here.
```
mgold: "Shrinkers must be deterministic, since they do not have access to a random number generator. It's the generator part of the fuzzer that's meant to find rare edge cases; it's the shrinkers job to make the failures understandable." or similar.

src/Shrink.elm
```diff
@@ -8,6 +8,86 @@ module Shrink exposing
...
+and doesn't necessarily have them all committed to memory. That allows it to
+take less space (interesting since there may be quite a lot of elements).
+
+That LazyList should also provide a constant number of shrunk values (if it
+provided an infinite number of them, tests using it might hang indefinitely
+at the shrinking phase).
+
+Also, they should never produce the same value twice. Doing so may result in
+tests looping over wrong values that return endlessly, hanging or generating
+stack too deep kind of exceptions.
```
mgold: It's a bit more subtle than that. The *shrinking process* must never produce the same value twice, even if it's through several shrinks. What you have written means that a single LazyList must contain unique items (which actually isn't essential -- if the duplicate fails, we only see if the first time, and if it passes, it gets skipped both times).

mgold:
Yes, it's a second round of changes. But keep up the good work!

Bastes:
All addressed :) the last one I'm a little less sure of, don't hesitate to criticize :)

src/Shrink.elm
```diff
@@ -8,6 +8,96 @@ module Shrink exposing
...
+and doesn't necessarily have them all committed to memory. That allows it to
+take less space (interesting since there may be quite a lot of elements).
+
+That LazyList should also provide a constant number of shrunk values (if it
+provided an infinite number of them, tests using it might hang indefinitely
+at the shrinking phase).
+
+Also, the shrinking process should never produce the same value twice, even
+through several shrinking attempts. Doing so may result in tests looping over
+wrong values that return endlessly, hanging or generating stack too deep kind
```
harrysarson: > looping over wrong values that return endlessly,

I think "that return" can be removed here.

src/Shrink.elm
```diff
@@ -8,6 +8,96 @@ module Shrink exposing
...
+of exceptions.
+
+    -- so, for example, a shrinker producing these outputs from these inputs could
+    -- make a test loop indefinitely:
+    shrinker 123 == [ 456, ... ]
+    shrinker 456 == [ 789, ... ]
+    shrinker 789 == [ 123, ... ] -- here we re-produce our first value and loop
```
harrysarson: re-produce -> reproduce

src/Shrink.elm
Bastes: Yup :)

Bastes:
All addressed :)

Bastes:
Never hesitate to nitpick :p thanks for the pass too :)

Bastes:
Don't hesitate to merge either ;) I don't have that power.

drathier:
I'm quite late to the party, but I feel like we could simplify this quite a lot. 

Can we rename shrinkers to simplifiers?

Do we have to mention the concept of a lazy list (or rose tree)? 

Can we remove the example that is just an int? It doesn't seem that clear to me.

Can we make the Node unshrunk example even larger, and the simplified output even simpler? The more of a wow-factor we have, the more likely people are to write shrinkers for their fuzzed types. The after example could just be "Leaf -1" or something. Turns out our tree doesn't support leaves with negative values 🤷.

Can we simplify the comments about infinite loops etc. to "don't write simplifiers that simplify values in a circle, like case True -> False ; False -> True, because how would we know when to stop simplifying then? Just decide one one being simpler."

mgold:
> Can we rename shrinkers to simplifiers?

It's a breaking change, tracked separately in #36. We typically wait for several such changes to accumulate.

> Do we have to mention the concept of a lazy list (or rose tree)?

I think we have to mention laziness, but can get away not talking about rose trees.

> Can we simplify the comments [about many things?]

Yes, it's possible to edit this down, and I especially like the idea of a more impactful example. However, we started with insufficient documentation. Given the choice, I'd rather have too much than too little. Another option is to move the details into a guide at the end, like [we did with floating point comparisons](https://package.elm-lang.org/packages/elm-explorations/test/latest/Expect#guide-to-floating-point-comparison).

We should also mention at the very top of the file that you only need to understand and write shrinkers if you are using `Fuzz.custom`.

Bastes:
There is it :) have fun.

src/Shrink.elm
```diff
@@ -388,29 +382,23 @@ Takes a tuple of shrinkers and returns a shrinker of tuples.
...
-                )
-            )
-        )
+    Lazy.List.map3 (\a2 b2 c2 -> ( a2, b2, c2 )) (shrinkA a) (shrinkB b) (shrinkC c)
+        |> append (Lazy.List.map2 (\a2 b2 -> ( a2, b2, c )) (shrinkA a) (shrinkB b))
+        |> append (Lazy.List.map2 (\a2 c2 -> ( a2, b, c2 )) (shrinkA a) (shrinkC c))
+        |> append (Lazy.List.map2 (\b2 c2 -> ( a, b2, c2 )) (shrinkB b) (shrinkC c))
+        |> append (Lazy.List.map (\a2 -> ( a2, b, c )) (shrinkA a))
+        |> append (Lazy.List.map (\b2 -> ( a, b2, c )) (shrinkB b))
+        |> append (Lazy.List.map (\c1 -> ( a, b, c1 )) (shrinkC c))
```
mgold: Let's keep code changes out of a docs PR. I would also need to verify that this change respects argument order, and that you haven't turned something around somehow.

src/Shrink.elm
```diff
@@ -8,34 +8,43 @@ module Shrink exposing
 helper functions to help you construct shrinking strategies.
```
mgold: This sentence would be a good thing to revise.

src/Shrink.elm
```diff
@@ -8,34 +8,43 @@ module Shrink exposing
...
-Shrinkers help obtaining a simpler case that still fails. Say after shrinking
-we find we still fail the test with the value:
+@docs convert, keepIf, dropIf, merge, map, andMap
+
+
+## What are Shrinkers and why do we need them?
 
-    9
+Shrinking is a way to try and find the "smallest" example that fails, in order
+to give the tester better feedback on what went wrong.
```
mgold: Between here and the very top of the module, we should have something to the effect of: "Shrinking is part of fuzzing, and the provided fuzzers have shrinkers already built into them. You really only have to write your own shrinkers if you use `Fuzz.custom`."

src/Shrink.elm
```diff
@@ -8,20 +8,104 @@ module Shrink exposing
...
+you're testing with.
+
+Let's say I'm writing a Shrinker for binary trees :
+
+    -- randomly-generated binary trees might soon become unreadable
+    type Tree a
+        = Node (Tree a) (Tree a)
+        | Leaf a
+
+In this context, "smaller" should probably mean "with less nodes", to make the
```
drathier: Fewer nodes

src/Shrink.elm
```diff
@@ -8,20 +8,104 @@ module Shrink exposing
...
+reason the test failed more obvious:
+
+    -- shrinking from this failing input:
+    Node (Node (Node (Leaf 8) (Node (Leaf -1) (Leaf 7))) (Node (Leaf 3))) (Leaf 4)
```
drathier: Up the numbers to multi-digit ones, especially the negative one. Possibly add even more `Node`s.

src/Shrink.elm
```diff
@@ -8,20 +8,104 @@ module Shrink exposing
...
+    -- to this failing input:
+    Leaf -1
+    -- it will probably take less time for the developer inspecting the failing
+    -- input to understand what went wrong (probably when there's a `Leaf -1` in
+    -- the tree)
+
+
+### How does shrinking work?
+
+When a test fails with a random value, this value is fed to the shrinker who
```
drathier: Who -> which

src/Shrink.elm
```diff
@@ -8,20 +8,104 @@ module Shrink exposing
...
+returns a LazyList of potentially failing inputs. Those will be passed to the
```
drathier: Do we have to mention LazyList? Isn't "a bunch of" better than "a LazyList of" here? Is it important to the reader here that it's a LazyList?

drathier: I find the "Those..." sentence a bit unclear. Can we link it better to the sentence before it (I feel like we lost track of the inputs we're talking about, because we had to specify which inputs we're speaking of, as if there were others), and reword it a bit?

src/Shrink.elm
```diff
@@ -8,20 +8,104 @@ module Shrink exposing
...
+test that failed given that original failing input.
+
+The fuzz test is rerun with the first "smaller" value. If the test now passes,
```
drathier: Is it relevant that it's the first smaller value? How about "one of the shrunken values", "some smaller value". Maybe there's a way to rewrite this section to make it clearer. Would a numbered list be easier to understand?

src/Shrink.elm
```diff
@@ -8,20 +8,104 @@ module Shrink exposing
...
+recursively shrunk, until no "smaller" value can be found.
+
+Once we can't find anymore "smaller" values that make the test fail, the last we
+found thus far is presented as the "smallest" failing value. If none has been
+found, the original failing value is presented instead.
+
+
+### How do I make my own Shrinkers?
+
+Shrinkers must be deterministic, since they do not have access to a random
```
drathier: Must be -> are. Must be sounds like it's a requirement that the developer has to keep in mind, but in reality they can't even make it nondeterministic if they wanted to.

src/Shrink.elm
```diff
@@ -8,20 +8,104 @@ module Shrink exposing
...
+Shrinkers have to return a LazyList, something that works a bit like a list.
+That LazyList may or may not have another element each time we ask for one,
+and doesn't necessarily have them all committed to memory. That allows it to
+take less space (interesting since there may be quite a lot of elements).
+
+That LazyList should also provide a finite number of shrunk values (if it
+provided an infinite number of them, tests using it might hang indefinitely
+at the shrinking phase).
+
+Shrinkers must never shrink values in a circle, like:
```
drathier: This is a special case of infinite LazyLists. Can we pose it as an example of the previous requirement, rather than as a new one? e.g. "LazyLists cannot be infinite, because then we'd have to spend an infinite time shrinking. For example, shrinking values in a circle:"

Also, shrinkers don't really hang, they shrink forever. Hang isn't as descriptive here.

src/Shrink.elm
```diff
@@ -33,18 +117,21 @@ import String
...
 
 
 {-| Perform shrinking. Takes a predicate that returns `True` if you want
-shrinking to continue (e.g. the test failed). Also takes a shrinker and a value
-to shrink. It returns the shrunken value, or the input value if no shrunken
+shrinking to continue (most likely the failing test for which we are attempting
+to shrink the value). Also takes the shrinker and the value to shrink.
+
+It returns the shrunken value, or the input value if there are no shrunken
 values that satisfy the predicate are found.
```
drathier: "It returns the shrunken value, or the input value if no shrunken
values that satisfy the predicate are found."

Bastes:
(oooh, that's a lot of fun ^^° ; I'll try to work around this asap)

src/Shrink.elm
Bastes: Aha, sorry, that was completely accidental (I tweaked with that to understand how it worked better, and forgot to do a `git add --patch` ^^°).

src/Shrink.elm
Bastes: Thanks :)

src/Shrink.elm
Bastes: Thanks :) (always have a hard time remembering when it's less vs when it's fewer ;p)

src/Shrink.elm
Bastes: Incoming a version that might go overboard (the tree is developed since it was overflowing the 80 chars limit quite a bit).

src/Shrink.elm
Bastes: Not sure about the numbered list. I'm trying something with bullet points to make the condition in the middle more obvious.

src/Shrink.elm
Bastes: Thanks :)

src/Shrink.elm
Bastes: Indeed :)

src/Shrink.elm
Bastes: I'm not sure what you meant here ^^° aren't those lists finite?

```Elm
[ True ]
[ False ]
```

As far as I understand, an infinite `LazyList Bool` for example would look like this:
```Elm
Lazy.List.cycle (Lazy.List.fromList [True, False])
```

I wanted to point the difference between returning an infinite `LazyList` and looping through the shrinking while still returning finite lists.

I'm still interested in a suggestion to make it better of course :)

Bastes:
Not to sound too impatient but it's been a week :) any other changes requested or do we merge this?

src/Shrink.elm
```diff
@@ -4,24 +4,145 @@ module Shrink exposing
...
+  - when the test passes, we discard the smaller value we tried it with and try
+    another one in the list
+  - when the test still fails, the rest of the list is discarded and that new
+    smaller failing value is recursively shrunk, until no smaller failing value
+    can be found
+
+Once we can't find anymore smaller failing value, the last we found thus far is
+presented as the smallest failing value. If none has been found, the original
+failing value is presented instead.
+
```
drathier: I tried to write down another version of this, but I still think we can do better. Not saying you should take mine over yours, I just wanted to write down roughly what I'd like to see in the docs. Wondering if we can hide `LazyList` from the user completely in a later version.

-----
<short intro>
A shrinker takes a value and returns a short list of smaller values. Once elm-test finds a failing fuzz test, it tries to shrink the input using the shrinker. We'll then try the smaller values as inputs to that fuzz test. If one of the smaller values also fail, we continue shrinking from there instead. Once the shrinker says that there are no smaller values, or no smaller values fail the fuzz test, we stop shrinking. 

<smaller vs simpler + infinite shrinking>
It's helpful to think of Shrinkers as returning simpler values rather than smaller values. For example, `1` is smaller/simpler than `47142`, and `-1` is smaller/simpler than `-47142`. Whether or not the shrunken value is actually smaller isn't that important, as long as we aren't shrinking in a loop. The `bool` shrinker shrinks `False` to `True`, but not vice versa. If it did, and your test failed no matter if this variable was `True` or `False`, there would always be a smaller/simpler value, so we'd never stop shrinking! We would just re-test the same values over and over again, forever!

### How do I make my own Shrinkers?

Fuzzers consist of two parts; a Generator and a Shrinker. The Generator takes a random Seed as input and returns a random value of your type, based on the Seed. Then, if the test fails on the random input, the shrinker takes that value and makes it smaller/simpler for you so you can guess what property of the input caused the test to fail.

src/Shrink.elm
drathier: I'm looking for some simpler way to describe this, that's all. If I read what looks like an explicitly described edge case, I assume it to be an edge case, and not a special case of something described before. If it matches with what I though before, I'd start thinking of other ways I could've read it which would make that an edge case, and if I can't find that either, I'll be confused. I think we can rephrase to avoid that confusion, I just don't know how.

mgold:
I'm going to go ahead and approve. Let's get @drathier's last round of changes in.

Bastes:
Thanks for the reply ; will be working on it as soon as I've got a minute :)

src/Shrink.elm
```diff
@@ -4,24 +4,145 @@ module Shrink exposing
...
+Let's say I'm writing a Shrinker for binary trees :
+
+    -- randomly-generated binary trees might soon become unreadable
+    type Tree a
+        = Node (Tree a) (Tree a)
+        | Leaf a
+
+In this context, "smaller" should probably mean "with fewer nodes", to make the
+reason the test failed more obvious:
```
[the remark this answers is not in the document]
Bastes: I reworked All the explanations there trying to uphold the storytelling all the way down.

src/Shrink.elm
```diff
@@ -4,24 +4,145 @@ module Shrink exposing
...
+          ( Leaf 999111 )
+          ( Leaf -148148 )
+        )
+      )
+    )
+    -- to this failing input:
+    Leaf -1
+    -- it will probably take less time for the developer inspecting the failing
+    -- input to understand what went wrong (probably when there's a `Leaf -1` in
+    -- the tree)
```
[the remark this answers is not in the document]
Bastes: I had trouble rephrasing this one following that idea ; I tried using another one. Feel free to suggest better :)

src/Shrink.elm
```diff
@@ -4,24 +4,145 @@ module Shrink exposing
...
+      ( Node
+        ( Leaf 123 )
+        ( Node
+          ( Leaf 999111 )
+          ( Leaf -148148 )
+        )
+      )
+    )
+    -- to this failing input:
+    Leaf -1
```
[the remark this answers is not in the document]
Bastes: Thanks :)

src/Shrink.elm
Bastes: Thanks for the rewording :) I've tried to find where they might fit, tell me what you think :)

src/Shrink.elm
```diff
@@ -57,48 +57,52 @@ Let's say I'm writing a Shrinker for binary trees :
...
+
+Now let's say we shrink it to the following simpler value:
+
     Leaf -1
-    -- it will probably take less time for the developer inspecting the failing
-    -- input to understand what went wrong (probably when there's a `Leaf -1` in
-    -- the tree)
+
+It will probably take less time for the developer inspecting the failing input
+to understand what went wrong. Next, the developer could check wether a negative
```
drathier: `whether`*

Bastes:
I've had a minute. Hope you like it ;)

src/Shrink.elm
Bastes: Looks like I forgot a castrated goat somewhere ;p

src/Shrink.elm
drathier: It's definitely good enough :)

src/Shrink.elm
```diff
@@ -4,24 +4,145 @@ module Shrink exposing
...
+values to test against.
+
+
+### What is "small" (or "simple")?
+
+That's kind of arbitrary, and depends on what kind of values you're fuzzing.
+When you write your own Shrinker, you decide what is small for the kind of data
+you're testing with.
+
+Let's say I'm writing a Shrinker for binary trees :
```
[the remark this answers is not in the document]
Bastes: That's french typing conventions for you I'm afraid ;p

src/Shrink.elm
```diff
@@ -4,24 +4,162 @@ module Shrink exposing
...
+                    (Leaf 999111)
+                    (Leaf -148148)
+                )
+            )
+        )
+
+This is a pretty big tree, there are many nodes and leaves, and it's difficult
+to tell which is responsible for the failing.
+
+Now let's say we shrink it to the following simpler value:
```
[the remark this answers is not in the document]
Bastes: I've reworked the sentence a bit differently ; the idea here is to say: "here's what it would be like without shrinking, and here's what it is like with shrinking".

Bastes:
All done.

drathier:
All done!

Bastes:
Yay, thanks guys :)
👍

PR #39, opened 2018-10-08
Fix links to package.elm-lang.org

rl-king:
* use https
* elm-explorations/elm-test -> elm-explorations/test
* link describe to `#describe` instead of `#test`


mgold:
Travis is complaining that `src/Expect.elm` is not formatted with elm-format, but this PR doesn't touch that file.

rl-king:
Might be due to elm-format 0.8.1 (released last week) being used by travis now instead of 0.8.0.

Push here, or separate PR? (It's one line of whitespace 🙂)

mgold:
If you want to reformat the codebase with the latest elm-format, that would be great. (Might be worth a larger discussion on pegging the elm-format version in CI; when a commit passes CI I expect it to pass a week later!)

PR #41, opened 2018-10-20
pull in elm-html-test (and elm-html-in-elm and elm-html-query)

avh4:


rtfeldman:
Just wanted to comment - great work on this @avh4! 😻  😻  😻

avh4:
@rtfeldman I think this is ready to merge if you want to look it over.

tests/test.sh
```diff
@@ -23,6 +23,6 @@ cp ./versions.dat elm_home/0.19.0/package/versions.dat
 
 export ELM_HOME="$(pwd)"/elm_home
 
-elm make src/Main.elm --output elm.js
+elm make --debug src/Main.elm --output elm.js
```
rtfeldman: `npm test` still passes for me if I remove the `--debug` flag here. Can we remove it?

As I understand it, we can't run tests with `--optimize` because we need the `Debug` module, but we can use that module whether or not we have the `--debug` flag. I believe the `--debug` flags only impacts whether the debugger gets compiled in.

src/Test/Html/Query.elm
```diff
@@ -0,0 +1,498 @@
...
+            Internal.contains
+                (List.map Inert.toElmHtml expectedElmHtml)
+                query
+                |> failWithQuery showTrace "Query.contains" query
+
+        Err errors ->
+            Expect.fail <|
+                String.join "\n" <|
+                    List.concat
+                        [ [ "Internal Error: failed to decode the virtual dom.  Please report this at <https://github.com/elm-explorations/test/issues>." ]
```
drathier: This error message has a link to where the user can report the bug. Could we get that link in all the other "please report this bug" messages?

src/Test/Html/Query/Internal.elm
```diff
@@ -0,0 +1,606 @@
...
+type Query msg
+    = Query (Inert.Node msg) (List SelectorQuery)
+    | InternalError String
+
+
+type SelectorQuery
+    = Find (List Selector)
+    | FindAll (List Selector)
+    | Children (List Selector)
+      -- First and Index are separate so we can report Query.first in error messages
```
drathier: Love this comment. The attention to detail is gold.

src/Test/Html/Query/Internal.elm
```diff
@@ -0,0 +1,606 @@
...
+queryErrorToString query error =
+    case error of
+        NoResultsForSingle queryName ->
+            queryName ++ " always expects to find 1 element, but it found 0 instead."
+
+        MultipleResultsForSingle queryName resultCount ->
+            queryName
+                ++ " always expects to find 1 element, but it found "
+                ++ String.fromInt resultCount
+                ++ " instead.\n\n\nHINT: If you actually expected "
```
drathier: Love the hint!

tests/test.sh
avh4: oh, so dev mode still includes the toString stuff?  If so, then removing `--debug` is fine
```suggestion
elm make src/Main.elm --output elm.js
```

avh4:
@drathier I believe all the changes you suggested are on code that was copied unchanged from other packages.  I agree with them all, but think those would be best as a separate PR.  Feel free to make one if you want!

drathier:
I (most likely) won't do a PR to fix those comments. It's up for grabs for anyone else 🙌

avh4:
> I (most likely) won't do a PR to fix those comments. It's up for grabs for anyone else 🙌

will add it to my list for a future twitch stream :)

rtfeldman:
Published as `1.2.0`!

Thanks again @avh4! 🏆

PR #43, opened 2018-10-30
Remove note on random not being from core

Y0hy0h:
The docs of the `custom` fuzzer method seem to be outdated. I did not find anything with "random" in the [core package](https://package.elm-lang.org/packages/elm/core/latest/).

Therefore I propose to remove that sentence.


Y0hy0h:
#44 is a similar PR. It will probably conflict with this one.

If you'd like to, just merge one and I can resolve the conflict on the other.

Y0hy0h:
Closed in favor of #45.

PR #46, opened 2018-10-31
Remove Expect.true and Expect.false

drathier:
These are the same changes as we merged in https://github.com/elm-community/elm-test/pull/242.


rtfeldman (approved):

mgold:
So... do we have a MAJOR change on master now?

rtfeldman:
Yep! I think that's fine...if we need to do a minor or patch release in the interim, I'm fine with branching off `1.0.0` just for that.

PR #50, opened 2018-11-14
Fix CI build after bump to 1.2.0

avh4:
Ideally, we should look at automating this to grab the value from `/elm.json` instead of having to update this manually everywhere, but this fixes it for now.


rtfeldman:
Looks good, thanks @avh4!

PR #71, opened 2019-01-25
In test script, pull package version from elm.json

mgold:
This prevents the tests from breaking whenever we bump the version.


mgold:
I didn't notice that with 1.2.1. It's a Haskell IO so it should be possible to read the JSON file "and then" use that to fill in the right version.

PR #84, opened 2019-04-09
Remove Simplify.map and Simplify.andMap

### LINKED ISSUE #65: Documentation still uses function-as-infix
This was removed in 0.19, should be updated. https://github.com/elm-explorations/test/blob/1.2.0/src/Shrink.elm#L527-L532

mgold:
These functions operate on LazyList, which we are hiding (so this
finishes the work mostly done in #79). Grepping for "lazy" in the
generated docs JSON finds no matches now, so I think there are not any
more that I've missed.

It's impossible to implement these functions (sensibly) if they were to
accept simplifiers instead of lazy lists. A simplifier is a function,
and a value of the type of the argument isn't available.

By deleting the relevant documentation, this closes #65.

Tagging major release blocker for obvious reasons.


drathier:
Looks good to me :)

Issue #48, opened 2018-10-31
Rename tuple and tuple3 into pair and triple?


drathier:
Since we only have two sizes of tuples, I feel like we could see if we can give the two distinct names. Thoughts?

mgold:
I really like this idea, and I hope I'm not premature in applying the major release blocker tag since it will be a breaking change.

mgold:
If we're doing a minor release for some reason, we can add these and keep the old names as deprecated, to give people time to change over.

drathier:
Assuming you meant `major release`?

rtfeldman:
I'm mostly kidding, but every time I think "threeple" I smile. 😄

drathier:
`twople` and `threeple` are perfect troll names to use in presentations 😂 Specially since `Twople` sounds fairly close to `tuple`, but not quite.

mgold:
@harrysarson Can you please make a PR with the rename? I don't think we should bother with aliases; just rename them.

harrysarson:
Thoughts on `tripple` vs `tripplet`?

PR #86, opened 2019-04-13
 Rename tuple and tuple3 functions to pair and triple. 

harrysarson:
See #48.

I have a slight preference for going with `triplet` over `triple`. This PR uses `triple` although it would be easy for me to change that if folk are keen.

### DIFF
--- a/tests/src/FuzzerTests.elm
+++ b/tests/src/FuzzerTests.elm
@@ -17,7 +17,7 @@ die =
 fuzzerTests : Test
 fuzzerTests =
     describe "Fuzzer methods that use Debug.crash don't call it"
-        [ describe "FuzzN (uses use replacemeForPair or replacmeForTriplet) testing string length properties"
+        [ describe "FuzzN (uses use pair or replacmeForTriplet) testing string length properties"
             [ fuzz2 string string "fuzz2" <|
                 \a b ->
                     testStringLengthIsPreserved [ a, b ]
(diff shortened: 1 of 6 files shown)


harrysarson:
When reviewing please double check that I have spelt triple correctly throughout this patch as I had a major battle with dyslexia whilst putting this together and am not entirely confident I won.

tests/src/FuzzerTests.elm
```diff
@@ -17,7 +17,7 @@ die =
 fuzzerTests : Test
 fuzzerTests =
     describe "Fuzzer methods that use Debug.crash don't call it"
-        [ describe "FuzzN (uses tupleN) testing string length properties"
+        [ describe "FuzzN (uses use pair or replacmeForTriplet) testing string length properties"
```
mgold: `(uses pair and triple)`

drathier:
`triple` is what I'd reach for right now, but `triplet` is more correct and my habits will change over time, so I'd be happy with `triplet`. `triple` means three times something, and a `triplet` is three closely related things.

harrysarson:
Googling suggests trio or triple.

https://english.stackexchange.com/a/39825

drathier:
Good point, I now prefer `triple` again. https://english.stackexchange.com/questions/39824/is-triple-the-proper-counterpart-of-pair-when-describing-a-group-of-three-item/39825#comment929406_337695

harrysarson:
Let's go with triple then

mgold:
Sure, I just need the one review comment addressed before merging.

tests/src/FuzzerTests.elm
harrysarson: Done!

harrysarson:
@mgold fixed it!

You may want to squash when you merge - if only to fix the spelling of triple in the commit messages!

PR #87, opened 2019-04-15
Fix float comparison error message

dbrgn:
I was at first a bit confused because I didn't find docs for `Float.within` :slightly_smiling_face:

### DIFF
--- a/src/Expect.elm
+++ b/src/Expect.elm
@@ -739,10 +739,10 @@ equateWith reason comparison b a =
 
         floatError =
             if String.contains reason "not" then
-                "Do not use Expect.notEqual with floats. Use Float.notWithin instead."
+                "Do not use Expect.notEqual with floats. Use Expect.notWithin instead."
 
             else
-                "Do not use Expect.equal with floats. Use Float.within instead."
+                "Do not use Expect.equal with floats. Use Expect.within instead."
     in
     if usesFloats then
         fail floatError


src/Expect.elm
```diff
@@ -313,6 +313,7 @@ minor inaccuracies introduced by floating point arithmetic.
     -- Fails because 0.1 + 0.2 == 0.30000000000000004 (0.1 is non-terminating in base 2)
     0.1 + 0.2 |> Expect.equal 0.3
 
+
```
harrysarson: This extra line of white space is causing elm format to fail verification, if you remove it (or run elm-format) then CI should go green :)

dbrgn: Oh sorry, I didn't notice that!

dbrgn: Hm, actually this is the result of running `elm-format 0.19` on the file. Maybe you're still using 0.18 in CI?

harrysarson: which version of elm-format are you running locally?

dbrgn: elm-format-0.19 0.8.0. I saw that there's a patch release available, might be that this changed the behavior.

I simply remove the additional line. Should pass the tests now.

harrysarson: Yeah, elm-format-0.19 0.8.1 is used in CI. Nice 1.

mgold:
Thanks!

mgold:
Oops, not docs, error messages.

PR #92, opened 2019-05-01
Make Fuzz.string generate unicode strings.

drathier:
Copy of https://github.com/elm-community/elm-test/pull/204. This PR was originally made for elm 0.18 back in August 2017. We delayed merging it since there was a bug with `String.reverse` where it didn't handle emojis or composite characters correctly. The bug has since been both fixed and published, so all 0.19 elm/core versions now reverse the strings we use in this fuzzer properly. There are still issues with the string reversal algorithm used in elm, but this new fuzzer doesn't generate strings that hit those cases.

### DIFF
--- a/src/Fuzz.elm
+++ b/src/Fuzz.elm
@@ -321,22 +321,19 @@ Shorter strings are more common, especially the empty string.
 string : Fuzzer String
 string =
     let
-        unicodeGenerator : Generator String
+        ( firstFreq, restFreqs ) =
+            unicodeCharGeneratorFrequencies
+
+        lengthGenerator =
+            Random.frequency
+                ( 3, Random.int 1 10 )
+                [ ( 0.2, Random.constant 0 )
+                , ( 1, Random.int 11 50 )
+                , ( 1, Random.int 50 1000 )
+                ]
+
         unicodeGenerator =
-            let
-                ( freq, rest ) =
-                    unicodeCharGeneratorFrequencies
-            in
-            frequencyList
-                (Random.frequency
-                    ( 3, Random.int 1 10 )
-                    [ ( 0.2, Random.constant 0 )
-                    , ( 1, Random.int 11 50 )
-                    , ( 1, Random.int 50 1000 )
-                    ]
-                )
-                freq
-                rest
+            frequencyList lengthGenerator firstFreq restFreqs
                 |> Random.map String.fromList
     in
     custom unicodeGenerator Simplify.string
--- a/src/Fuzz.elm
+++ b/src/Fuzz.elm
@@ -48,6 +48,7 @@ import Fuzz.Internal as Internal
         , Valid
         , ValidFuzzer
         , combineValid
+        , frequencyList
         , invalidReason
         )
 import Lazy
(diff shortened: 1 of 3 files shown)


drathier:
N.B. master already has a major change on it, renaming the exposed Shrink module to Simplify.

tests/src/FuzzerTests.elm
```diff
@@ -287,3 +288,61 @@ whitespace =
...
+
+
+unicodeStringFuzzerTests : Test
+unicodeStringFuzzerTests =
+    describe "unicode string fuzzer" <|
+        [ test "generates ascii" <|
+            \() ->
+                expectTestToFail <|
+                    fuzz string "generates ascii" <|
+                        \str -> str |> String.contains "E" |> Expect.equal True
```
mgold: I'm confused by these tests. If the fuzzer is working correctly, at least one in 100 strings will contain `"E"`. What you're asserting is that at least one in 100 strings does *not* contain the substring.

I think you need to change `True` to `False`, asserting that **no** generated string contains `"E"`. `expectToFail` flips this to **at least one** generated string contains it.

drathier:
Ready for another round of reviews. @mgold

drathier:
I think it can. Do we want to limit ourselves to just valid unicode strings? I bet for 99% of cases it would shrink down to a single one.

harrysarson:
https://en.m.wikipedia.org/wiki/Combining_character

> In Unicode, diacritics are always added after the main character (in contrast to some older combining character sets such as ANSEL), so it is possible to add several diacritics to the same character, although as of 2010, few applications support correct rendering of such combinations.

So multiple combination diacritics are valid!

drathier:
If there's a bug that requires more than one diacritic mark in a row to trigger, it doesn't hurt if we find it, even if the diff is hard to read (hex output in diff helps). 

If it's triggered by multiple combining characters in a row but doesn't require those exact characters, it'll shrink to something readable.

I think we should keep it as-is, mainly on the argument that it's easier to code/test/understand.

drathier:
@mgold Are you thinking of this multi-line import exposing? Github folded it a bit unhelpfully. https://github.com/elm-explorations/test/pull/92/files#diff-adac3ec51c96ebc8ace1960bee9f3d35R51 There's an addition to the public interface of Fuzz.Internal, but that module isn't exposed, and I need that function in Fuzz.elm.

src/Fuzz.elm
```diff
@@ -280,41 +281,65 @@ asciiCharGenerator =
...
+                    unicodeCharGeneratorFrequencies
+            in
+            frequencyList
+                (Random.frequency
+                    ( 3, Random.int 1 10 )
+                    [ ( 0.2, Random.constant 0 )
+                    , ( 1, Random.int 11 50 )
+                    , ( 1, Random.int 50 1000 )
+                    ]
+                )
```
harrysarson: I found this function slightly hard to grok and, whilst having a play with it to get my head round what the code does, rewrote it slightly. I believe the function might be slightly easier to read is this form:

```elm
string : Fuzzer String
string =
    let
        ( firstFreq, restFreqs ) =
            unicodeCharGeneratorFrequencies

        lengthGenerator =
            (Random.frequency
                ( 3, Random.int 1 10 )
                [ ( 0.2, Random.constant 0 )
                , ( 1, Random.int 11 50 )
                , ( 1, Random.int 50 1000 )
                ]
            )

        unicodeGenerator =
            frequencyList lengthGenerator firstFreq restFreqs
                |> Random.map String.fromList
    in
    custom unicodeGenerator Simplify.string 
```

drathier: Looks good to me :) Took it as you wrote it and ran it though elm-format.

mgold:
@drathier Ah, that's an import, not an exposing. Looks good!

PR #130, opened 2020-03-24
Correct math mistake

ursi:
The most common convention in mathematics is that `0` has no sign. The numbers greater than or equal to `0` are generally described with the term "non-negative".

### DIFF
--- a/src/Test.elm
+++ b/src/Test.elm
@@ -268,7 +268,7 @@ The number of times to run each fuzz test. (Default is 100.)
     import Expect
 
 
-    fuzzWith { runs = 350 } (list int) "List.length should always be positive" <|
+    fuzzWith { runs = 350 } (list int) "List.length should never be negative" <|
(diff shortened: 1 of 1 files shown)


harrysarson:
This does make the example more correct. :+1:

drathier:
How about `List.length should never be less than 0`? :) I'm personally trying to avoid double-negatives in my docs, since in some spoken languages (e.g. Spanish) `not not x` = `not x`.

ursi:
I'm very comfortable with the term `non-negative` because I've been exposed to it so much, but if you don't like it, I think what you just suggested works just as well 👍

harrysarson:
non-negative is a very standard maths saying. My favourite is `List.length should never be negative`.

drathier:
Yep, I like that one too, and the three of us seem to agree

drathier:
Thanks!

PR #143, opened 2020-06-20
Fix CI and update to elm 0.19.1

harrysarson:
- fix CI in new node versions by bumping deps
- simplify ci config
- ~~upgrade to elm 0.19.1~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ elm-stuff/
 docs/
 *.html
 /documentation.json
+/package-lock.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
-sudo: false
-
 language: node_js
+
 node_js:
   - "8"
   - "10"
@@ -10,10 +9,3 @@ os:
   - osx
   - linux
 
-install:
-  - node --version
-  - npm --version
-  - npm install
-
-script:
-  - npm test
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
   "author": "Richard Feldman",
   "license": "BSD-3-Clause",
   "devDependencies": {
-    "elm": "^0.19.0-bugfix2",
-    "elm-format": "^0.8.1"
+    "elm": "latest-0.19.0",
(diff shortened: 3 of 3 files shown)


harrysarson:
We could just pick the first commit (the second two are more controversial)

harrysarson:
I do not understand:

```
elm-format --validate src tests/src benchmarks

{

  "name": "elm-test-tests",

  "version": "0.0.0",

  "description": "tests for elm-test, so you can elm-test while you elm-test",

  "scripts": {

    "test": "cat package.json && elm --version && elm make --docs=documentation.json && (cd tests && ./test.sh) && elm-format --validate src tests/src benchmarks"

  },

  "author": "Richard Feldman",

  "license": "BSD-3-Clause",

  "devDependencies": {

    "elm": "latest-0.19.0",

    "elm-format": "latest-0.19.0"

  },

  "dependencies": {}

}

0.19.1

...
```

harrysarson:
Okay, I tracked the issue with failing CI down. The problem is that node_modules are cached and therefore the wrong version of elm is used. The solution is to delete the cache, although someone with commit access to this repository will need to do so (see image).

image

---

Tagging @rtfeldman and @mgold as folk I think have commit access here.

drathier:
I have commit access. Nuked the caches and rebuilt, it's green now :)

harrysarson:
Thanks! Prior to this change we only ran of x86 Linux and x86 macos. Those platforms will have identical floating point behaviour (or so I believe as they both have 64 bit intel processors)

drathier:
Iirc we had exact matching float tests for expectWithin at first, but it failed in CI. Then I ran the tests locally on Windows and Mac and they differed. Afaik architecture, cpu generation and OS can all affect float calculations, so I think we should test broadly.

harrysarson:
@drathier you have convinced me that most my changes to .travis.yml were misguided. I have rebased and now only the redundant parts of the ci config have been changed.

This is good to go from my point of view.

drathier:
All good then :) Thanks for helping out!

PR #145, opened 2020-06-30
update compiler to 0.19.1

harrysarson:
Upgrading requires a hack to work around compiler error when importing non-local kernel models.

I am not sure it is worth it.

### DIFF
--- a/tests/make.sh
+++ b/tests/make.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -eu
+
+sed -i 's/import Elm\.Kernel\.Debug/-- import Elm\.Kernel\.Debug/g' src/Test/Internal.elm
+
+elm make --docs=documentation.json
+exit_code=$?
+
+sed -i 's/-- import Elm\.Kernel\.Debug/import Elm\.Kernel\.Debug/g' src/Test/Internal.elm
+
+exit $exit_code
--- a/tests/make.sh
+++ b/tests/make.sh
@@ -2,11 +2,19 @@
 
 set -eu
 
-sed -i 's/import Elm\.Kernel\.Debug/-- import Elm\.Kernel\.Debug/g' src/Test/Internal.elm
+# Check that this script is run from top level directory
+if [[ $(git rev-parse --show-toplevel) != $(pwd) ]]; then
+  >&2 echo "Please run ./tests/make.sh from git root directory!"
+  exit 1
+fi
+
+FILE_WITH_IMPORT=src/Test/Internal.elm
+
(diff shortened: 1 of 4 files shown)


harrysarson:
@drathier can you purge the cache please :angel:

drathier:
Done, and restarted

harrysarson:
Test failure is legit, some sort of sub par version of sed on mac.

drathier:
LMK when you want a review and/or merge

harrysarson:
I would love a review! I think this is the _best_ patch out there to upgrade the compiler to 0.19.1. However, tests/make.sh contains such a nasty collection of hacks that it may be better to keep using elm 0.19.0.

tests/make.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Check that this script is run from top level directory
+if [[ $(git rev-parse --show-toplevel) != $(pwd) ]]; then
+  >&2 echo "Please run ./tests/make.sh from git root directory!"
+  exit 1
+fi
+
+FILE_WITH_IMPORT=src/Test/Internal.elm
+
+sed -i.bak 's/import Elm\.Kernel\.Debug/-- import Elm\.Kernel\.Debug/g' "$FILE_WITH_IMPORT"
```
drathier, line 13: Is this workaround for this bug? https://github.com/elm/compiler/issues/2123

harrysarson: Yes, the other work around would be to permanently delete this import. However, I think it is better to leave the import there as it self documents the use of kernel code.

drathier:
Do we expect to find/avoid any bugs/issues by moving to 0.19.1? It seems like a pretty hacky change. I haven't worked in this code base for a while, so I don't remember what we should be thinking about here.

harrysarson:
There is not much difference between elm 0.19.0 and 0.19.1 so unlikely to make much difference. Keeping our dependencies up to date is my main motivation. It also seems like a bit of an issue if the elm test runner's test suite cannot pass with the most recent release of elm.

drathier:
@rtfeldman @mgold any input?

rtfeldman:
I don't have strong feelings about this one way or the other...one the one hand, it's not like we're getting a big benefit, but on the other hand, it's two commits, so easy to roll back if we regret it.

I'd say feel free to merge if you like, @harrysarson!

harrysarson:
@drathier I can squash my commits before we merge this.

drathier:
Aight, imma merge this then :)

harrysarson:
> @drathier can you purge the cache please 👼

Thanks for merging! Please purge the cache for master!

drathier:
Thanks for your help! Appreciate the link too :) Caches are nuked, builds restarted, and master is green!

PR #148, opened 2020-07-06
Update `Expect.ok` and `Expect.err` documentation

marc136:
Hello, 
it seems like the description of `ok` was a direct copy of the description of `err`, so I swapped the words and updated the result.

Also, the `String.toInt` in elm 0.19 returns a `Maybe` (see [docs](https://package.elm-lang.org/packages/elm/core/1.0.0/Result#fromMaybe)), so I added a `Result.fromMaybe` to the examples.

If I should split it into two commits, please say so.


drathier:
Good catch, thanks! 

Maybe long-term finding a function that actually returns a Result for the docs is good, like a json decoder.

PR #153, opened 2020-10-03
Prefix internal Test type variants so node-test-runner can find them

lydell:
**Explanation for why node-test-runner needs this:** https://github.com/rtfeldman/node-test-runner/pull/442#issuecomment-713526271

(Original discussion leading to this PR: https://github.com/rtfeldman/node-test-runner/pull/442#discussion_r495579878)

### DIFF
--- a/src/Test/Internal.elm
+++ b/src/Test/Internal.elm
@@ -7,20 +7,24 @@ import Test.Expectation exposing (Expectation(..))
 import Test.Runner.Failure exposing (InvalidReason(..), Reason(..))
 
 
+{-| All variants of this type has the `ElmTestVariant__` prefix so that
+node-test-runner can recognize them in the compiled JavaScript. This lets us
+add more variants here without having to update the runner.
+-}
 type Test
-    = UnitTest (() -> List Expectation)
-    | FuzzTest (Random.Seed -> Int -> List Expectation)
-    | Labeled String Test
-    | Skipped Test
-    | Only Test
-    | Batch (List Test)
+    = ElmTestVariant__UnitTest (() -> List Expectation)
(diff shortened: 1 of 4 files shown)


harrysarson:
I really want this change, it will future proof a hack we need to make in the test runner to detect tests without parsing the .elmi files.
I think the ugliness is worth it.

drathier:
Yes, I want more context and documentation first! https://github.com/rtfeldman/node-test-runner/pull/442#issuecomment-713526271

src/Test/Internal.elm
```diff
@@ -7,20 +7,24 @@ import Test.Expectation exposing (Expectation(..))
 import Test.Runner.Failure exposing (InvalidReason(..), Reason(..))
 
 
+{-| All variants of this type has the `ElmTestVariant__` prefix so that
+node-test-runner can recognize them in the compiled JavaScript. This lets us
+add more variants here without having to update the runner.
+-}
```
drathier, line 13: Let's add a link to this PR as well so people looking at the Test type definition know what's up! https://github.com/elm-explorations/test/pull/153

drathier:
Just want that link, then I'd be happy to merge this.

src/Test/Internal.elm
lydell: Link added!

drathier: One more thing; CI wants you to elm-format it too, so CI goes green for non-linux builds :)

lydell: 🤦‍♂️ Fixed

drathier: I'll merge it when it goes green :)

drathier:
Cool, let's give it a go :)

Issue #144, opened 2020-06-22
Trivial failing test takes forever to run, slow output diffing might be reason.


malaire:
This test takes about 120 seconds to run even though it's trivial comparison of `( List String, Int )` with 312 `String`:s.

```
module Example exposing (suite)

import Expect exposing (Expectation)
import Test exposing (..)


suite : Test
suite =
    test "simple comparison" <|
        \_ -> input |> Expect.equal expected


input : ( List String, Int )
input =
    ( [ "8000000000000000"
      , "CF935BDAA77581F6"
      , "81771566D7D8CD95"
      , "4258409FE1920B68"
      , "F921484837F8AF8F"
      , "D6CEEBD584E9EB9A"
      , "A984EDF9DC4EA196"
      , "E6564BE6FFB5574D"
      , "8D565567B36A191F"
      , "7F07C14F85E4DCC1"
      , "59DAB54371E62887"
      , "42B78DA549FE3BD4"
      , "566887E83E35D231"
      , "945945EE5136EC56"
      , "CE2C10B3F5266E65"
      , "26989A0966A07BF6"
      , "421E0A8E70C2668C"
      , "5AE6F9A39869F350"
      , "E954BC1232A1F7E9"
      , "296EEA2ECE4943AF"
      , "DB52F14BEFB75E14"
      , "6E8B2FEF3B30EC34"
      , "7E2710688A70BFBC"
      , "16A2340A4A12874A"
      , "9091F8A8E4908E9F"
      , "E50AB3DAA1EAC654"
      , "43C021200F0E1876"
      , "9F854FA6AC18887C"
      , "5870740D3F706E1D"
      , "69388EFCB0846057"
      , "C7B3DAE8050BBF40"
      , "B08F671353A3D626"
      , "E62E9DCD31CB4FAB"
      , "8085EB7C397106E6"
      , "D436985A6AB8DD4F"
      , "9DB95F220770203F"
      , "06DDBC3E96F80BF9"
      , "58EF4A1FDE98F02E"
      , "F628A8D6E50EAF67"
      , "ED011E3C7BC234F9"
      , "69C5D08A54720F85"
      , "C0B0F7EF1E28E24A"
      , "20329E5C0BB45024"
      , "4054D71E3903C716"
      , "86883676F17EED07"
      , "BDB6325BBE1BA0B2"
      , "17B95D586476FE09"
      , "D867286208197DE0"
      , "C235F6B97F87633A"
[...]

PR #155, opened 2020-10-18
Skip diffing if input is too large

drathier:
Workaround for #144 . Not sure how to test this in CI; we would somehow need to measure the time it takes for CI to run a specific failing test, so I've commented out the tests for now.


drathier:
Tests fail to run elm-format on ubuntu because of missing glibc; `version `GLIBC_2.27' not found`. Does anyone know the fix?

harrysarson:
Looks good, yeah hard to think of a way to test. Maybe we run this test in a separate process and put a timeout on it. 

To fix elm-format in CI you have to opt in to a more recent ubuntu version `dist: focal`

PR #157, opened 2020-10-22
Curry pair and triple fuzzers.

drathier:
I think it would help composibility if we curried the pair and triple fuzzers. This PR does that.


Janiczek:
I would love that - I believe I have something similar in the WIP shrinking PR. The ergonomics seem better to me.

PR #159, opened 2020-11-02
Fix links, from elm-community to elm-explorations

cllns:
Just ran across this issue looking at the docs at: https://package.elm-lang.org/packages/elm-explorations/test/latest/Test#fuzz

When I clicked the `Fuzzer` link, it took me to `elm-community/test` which isn't maintained (or compatible with 0.19).

Seraching through the codebase, I looked at every instance of `elm-community`, and most are links to the archived repo (and would break if we tried to link them to this repo now).

I did find another one, where it says to create an issue in the archived repo, which is not possible. Now it says to create an issue in this repo.

### DIFF
--- a/src/Test.elm
+++ b/src/Test.elm
@@ -342,7 +342,7 @@ fuzzWithHelp options aTest =
 
 
 {-| Take a function that produces a test, and calls it several (usually 100) times, using a randomly-generated input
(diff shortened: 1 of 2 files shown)


harrysarson (approved):

drathier:
Good catch! There's a few more mentions of "elm-community" in this repo but I think they should stay as-is, since they are referring to the old repo's issues or the old repo changelog. https://github.com/elm-explorations/test/search?q=%22elm-community%22&type= in src/Tests.elm

Issue #164, opened 2021-04-05
The document of `custom` has a mistake.


Hexirp:
The description of `custom` has a mistake on its code snippet .

https://github.com/elm-explorations/test/blob/74076ea9869d1ac70c8c7f674c87946784911506/src/Fuzz.elm#L93-L118

It uses `Random.bool`, however [Random of elm/random 1.0.0 ](https://package.elm-lang.org/packages/elm/random/1.0.0/Random) doesn't contain `bool`.

Hexirp:
I suggest:

```elm
     type Question 
         = Name String 
         | Age Int 
  
     question = 
         let 
             generator = 
                 uniform True [False]
                     |> Random.andThen 
                         (\b -> 
                             if b then 
                                 Random.map Name genName 
  
                             else 
                                 Random.map Age (Random.int 0 120) 
                         ) 
  
             shrinker question = 
                 case question of 
                     Name n -> 
                         Shrink.string n |> Shrink.map Name 
  
                     Age i -> 
                         Shrink.int i |> Shrink.map Age 
         in 
         Fuzz.custom generator shrinker 
```

drathier:
Good catch! Wanna make a PR?

PR #166, opened 2021-04-07
Fix the code snippet in the document of Fuzz.custom

Hexirp:
It uses `Random.bool`, however [Random of elm/random 1.0.0 ](https://package.elm-lang.org/packages/elm/random/1.0.0/Random) doesn't contain `bool`.

See also: https://github.com/elm-explorations/test/issues/164


drathier:
LGTM, thanks for the fix!

PR #172, opened 2021-09-03
Add missing `|> Query.fromHtml` to doc example of `Selector.containing`

marc136:
This is just a minimal change where the documentation code example was missing `Query.fromHtml` and did not work because of that.

See https://package.elm-lang.org/packages/elm-explorations/test/1.2.2/Test-Html-Selector#containing


avh4 (approved):

rtfeldman (approved):
Thanks @marc136!

PR #173, opened 2021-09-08
Add a GitHub Action workflow for tests and formatting

mpizenberg:


mpizenberg:
This will probably need _an action_ (pun intended) on your part @harrysarson to activate the GH action. This may not work on the first shot so maybe tweaks will be needed too since I can't easily test the action setup.

harrysarson:
do you know how I activate it?

mpizenberg:
No idea sorry. It always was active for my projects by just creating the workflow files. And one time someone new contributed to a project of mine and I had to click on "run action" or something similar because it was the first time they contributed to the repo, but the action was already setup. So for Action setup I don't know. Maybe requires something by someone who owns elm-explorations?

avh4:
The actions config is always a bit confusing to me...  Note also that depending on the settings, there can be differences in which actions it runs based on who that branch author is and whether it's a PR or not.

I think the relevant settings are at Settings -> Actions, and check that both the "Actions permissions" section and "Fork pull request workflows from outside collaborators" seem to have settings that make sense.

mpizenberg:
@harrysarson did you find the relevant settings? who should we ask otherwise that has the rights to make changes?

harrysarson:
I tried merging, we can always fix it in a future mr

mpizenberg:
Well at least checking formatting worked ^^ https://github.com/elm-explorations/test/runs/3571514113

Issue #134, opened 2020-04-10
Test.Html.Query.has wrongfully returns a failed expectation even when the clauses in the Selector are there


sashaafm:
Hello, I'm seeing the following error in a test:
```

    ▼ Query.findAll [ tag "svg" ]

        1)  <svg class="some-class-type-1 some-class-icon v-align-middle" height="20" width="20">
                <circle cx="10" cy="10" fill="#fb8532" r="9" shape-rendering="geometricPrecision" stroke="black" stroke-width="0.5">
                </circle>
            </svg>

        2)  <svg class="some-class-type-2 some-class-icon v-align-middle" height="20" width="20">
                <circle cx="10" cy="10" fill="#f66a0a" r="9" shape-rendering="geometricPrecision" stroke="black" stroke-width="0.5">
                </circle>
            </svg>

        3)  <svg class="some-class-type-0 some-class-icon v-align-middle" height="20" width="20">
                <circle cx="10" cy="10" fill="#ffdf5d" r="9" shape-rendering="geometricPrecision" stroke="black" stroke-width="0.5">
                </circle>
            </svg>


    ▼ Query.index 0

        1)  <svg class="some-class-type-1 some-class-icon v-align-middle" height="20" width="20">
                <circle cx="10" cy="10" fill="#fb8532" r="9" shape-rendering="geometricPrecision" stroke="black" stroke-width="0.5">
                </circle>
            </svg>


    ▼ Query.has [ class "some-class-type-1" ]

    ✗ has class "some-class-type-1"
```

as you can see `Query.index 0` finds an `svg` with `some-class-type-1`, and I use a `Query.has [ class "some-class-type-1" ]`. However, it reports as not having that class `some-class-type-1`. AFAIK, this is a bug because `Query.has` is supposed to return a successful expectation if the `Selector` clauses given to it are true, which they seem to be in this case.

I'm providing the test code below:
```
model
|> view
|> Query.fromHtml
|> Query.findAll [ Selector.class "row" ]
|> Query.keep (Selector.tag "svg")
|> Expect.all
     [ Query.index 0
         >> Query.has [ Selector.class "some-class-type-1" ]
     , Query.index 1
         >> Query.has [ Selector.class "some-class-type-2"]
     , Query.index 2
         >> Query.has [ Selector.class "some-class-type-0"]
     ]
```

If there's any further information I can give please do tell. 

Thank you

TheManSpeaker:
I'm having the same issue:

```
[...]

PR #175, opened 2022-01-15
Ns class selectors

Confidenceman02:


Confidenceman02:
@rtfeldman fix as promised.

If I set my pedanto meter really high, I think there is a valid argument to move these selectors into their own `Svg.Html.Selector` namespace but perhaps it is not required.

I also noticed the `MakeTestRegistry.hs` elm package versions are not the latest and greatest. Would you like me to bump those? 

Keen to get your feedback on this implementation.

elm.json
```diff
@@ -21,6 +21,7 @@
         "elm/html": "1.0.0 <= v < 2.0.0",
         "elm/json": "1.0.0 <= v < 2.0.0",
         "elm/random": "1.0.0 <= v < 2.0.0",
+        "elm/svg": "1.0.1 <= v < 2.0.0",
```
rtfeldman, line 24: Can this lower bound be `1.0.0` instead of `1.0.1`?

tests/elm.json
```diff
@@ -6,20 +6,20 @@
...
             "elm/url": "1.0.0",
-            "elm/virtual-dom": "1.0.0",
+            "elm/virtual-dom": "1.0.2",
             "elm-explorations/linear-algebra": "1.0.3",
             "elm-explorations/markdown": "1.0.0",
             "elm-explorations/test": "1.2.2",
-            "elm-explorations/webgl": "1.0.1",
-            "jinjor/elm-diff": "1.0.5",
+            "elm-explorations/webgl": "1.1.3",
+            "jinjor/elm-diff": "1.0.6",
```
rtfeldman, line 22: I'd rather we not update this file if possible, just to make sure everything still works with the older versions - to minimize the chances that these changes cause regressions for anyone! 😄

rtfeldman:
> I think there is a valid argument to move these selectors into their own `Svg.Html.Selector` namespace

Given how much these selectors will overlap with the `Html` selectors (e.g. if I'm testing SVG things, I'll want to import the `Html` ones anyway) I think the current approach seems better.

@tesk9 What do you think? Also, any feedback on the PR in general?

rtfeldman:
> I also noticed the `MakeTestRegistry.hs` elm package versions are not the latest and greatest. Would you like me to bump those?

I'd prefer to keep this change as minimal as possible, so let's not!

elm.json
Confidenceman02: Done 👍

tests/elm.json
Confidenceman02: Done 👍

tests/src/Test/Html/SelectorTests.elm
```diff
@@ -16,6 +18,104 @@ all =
...
+                let
+                    svgClass =
+                        "some-NS-class"
+                in
+                Svg.svg
+                    [ SvgAttribs.class svgClass, SvgAttribs.class "another-NS-class" ]
+                    [ Svg.circle [ SvgAttribs.cx "50", SvgAttribs.cy "50", SvgAttribs.r "40" ] [] ]
+                    |> Query.fromHtml
+                    |> Query.has [ classesNS [ svgClass ] ]
+        , test "exactClassNameNS selector finds the exact class value on svg" <|
```
tesk9, line 74: I'm not sure that we should support `exactClassName` on SVGs. Might lead to more confusion with respect to bugs like https://ellie-app.com/gsphpMkJVN5a1 ?

tesk9 (commented):
Initially on reading through, I felt a little unsure about the API. I wasn't sure if users would get that `NS`-means-SVG, especially since the module name is `Test.Html.Selector`. I wouldn't necessarily expect to have svg-specific helpers in a module with `Html` in the name, you know? So I did a little exploration of what info we currently have, to see if I could suggest an alternate approach to changing the Selectors.

I printed out what is produced when running tests with
```
Svg.svg [ SvgAttribs.class svgClass ] [  ]
    |> Query.fromHtml
```
and
```
Html.div [ Attr.class someClass ] []
    |> Query.fromHtml
```

I found the results to be mostly the same, expect for the attribute name:

SVG node entry:
```
NodeEntry {
    children = [],
    descendantsCount = 1,
    facts = {
        attributeNamespace = Nothing,
        boolAttributes = Dict.fromList [],
        events = Dict.fromList [],
        stringAttributes = Dict.fromList [("class","some-NS-class")],
        styles = Dict.fromList []
    },
    tag = "svg"
    }
```

HTML node entry:
```
NodeEntry {
    children = [],
    descendantsCount = 0,
    facts = {
        attributeNamespace = Nothing,
        boolAttributes = Dict.fromList [],
        events = Dict.fromList [],
        stringAttributes = Dict.fromList [("className","some-class")],
        styles = Dict.fromList []
    },
    tag = "div"
    }
```

I think what this means is that we can change the Query inernal code instead of the Selector API code!

I'm making a PR off of this one. Let me know what you think!

Confidenceman02:
> Initially on reading through, I felt a little unsure about the API. I wasn't sure if users would get that `NS`-means-SVG, especially since the module name is `Test.Html.Selector`. I wouldn't necessarily expect to have svg-specific helpers in a module with `Html` in the name, you know? 

I had the same intuition and this is very easily fixed by treating these selectors as Svg specific. i.e. `Test.Svg.Html.Selector`.
I know @rtfeldman wasn't totally keen on this. Considering Svg's in elm are treated similarly, consumers of this package don't have to do too many mental gymnastics to accept if you are importing and testing `Svg.Html` or `Svg.Html.Attrbute` then logically you would find those selectors as `Test.Svg.Html.Selector`.

I think this is analogous to the Html packages and how people import and use them and outweighs the convenience of having them all live in `Test.Html.Selector`.

 However, it would likely be just the `class` selector that is problematic so I can see the argument against this.

> I found the results to be mostly the same, expect for the attribute name:
> 

> I think what this means is that we can change the Query internal code instead of the Selector API code!
> 
I 100% encourage exploration of this idea, but at the same time, here be dragons!

- Changing the internal query means that we would be asserting that a Html class selector works for both svg and html, which is not actually how Html works in elm. Setting a class on an svg element with a `Html.Attribute.class` throws an exception, and thus I would argue it should fail in Test also. Perhaps more verbosely than it currently does.

- From lessons learned in recent `elm-css` fixes, svg and html elements differ in how classes are set and we found that one solution does not cover both scenarios. Given this, I don't see how we can get around the fact you will need to query 'className' for html and `class` for svg. This means you would need some context about am I querying a svg element? Or a html element? From my understanding of the code that would be a decent refactor. This is just from my understanding of the code of coarse and not to be taken as gospel.

- A change in implementation to the currently working html queries seems far more invasive than exposing a class selector that maps to a specific svg query.

tests/src/Test/Html/SelectorTests.elm
Confidenceman02: Hmm.. I don't quite understand? Could you elaborate? `exactClassNameNS` simply queries the full class attribute on the svg which could consist of multiple classes, not sure how that pertains to the bug you linked.

tesk9:
I'm not opposed to API changes in general, but I would want people who worked on the dom portion of the package to be pulled in before anything major changes.

I also think that the key problem here doesn't necessarily require API changes to fix.

My thinking is the key problem is that there are 2 ways to set the `class`, and the elm test will only find classes set one these 2 ways. (I assert that it's a separate problem that there's an error if you try to set a className on an SVG node.)

I think a solution that doesn't fix these tests:

```
suite : Test
suite =
    describe "class attribute assertions"
        [ describe "Using Html.Attribute.class" <|
            classAssertions <|
                Html.div [ Attr.class "some-custom-class" ] []
        , describe "Using Html.Attribute.property" <|
            classAssertions <|
                Html.div [ Attr.property "className" (Json.Encode.string "some-custom-class") ] []
        , describe "Using Html.Attribute.attribute" <|
            classAssertions <|
                Html.div [ Attr.attribute "class" "some-custom-class" ] []
        ]


classAssertions : Html.Html msg -> List Test
classAssertions html =
    [ test "Test.Html.Selector.class finds the class" <|
        \_ ->
            html
                |> Query.fromHtml
                |> Query.has [ Selector.class "some-custom-class" ]
    , test "Test.Html.Selector.classes finds the class" <|
        \_ ->
            html
                |> Query.fromHtml
                |> Query.has [ Selector.classes [ "some-custom-class" ] ]
    ]
```
isn't solving the problem of the Query module losing track of HTML classes.

(It's not clear to me whether `exactClassName` has `className` in the name because it's supposed to be about the className property. Depending on what the intent is, it might make sense for it to be in the `classAssertions` list too. In my PR, I treated `exactClassName` as being about the method of setting the className. I could see an argument either way.)

I agree that SVGs are related to the problem (in that you can only set classes on them in the way that elm-test can't yet test for unless you want errors in the browser), but again I think the core problem is the attribute/property distinction.

I'll add these tests explicitly to my PR.

Confidenceman02:
It's a good point, I was viewing `exactClassName` as "the class name" not "the value of className property" even though weirdly, both are correct. I came to this thinking because the selectors relating to classes are `Selector.class` and `Selector.classes` respectively, not `Selector.className` or `Selector.classNames`.

Furthermore, if we leave out `exactClassName`, we essentially lose the ability to assert the entire class value i.e. "class1 class2 class3 class4" but only for `svg` which doesn't make intuitive sense in my opinion. My understanding is `exactClassName` works the same as the other class selectors, the only difference being it doesn't `String.split` the value.

Ultimately, I would rather leave the current working implementations of Html selectors alone and instead mimic those behaviours for NS elements specifically. I think this provides the least chance peoples tests will break due to something we don't know, and directly addresses issue #134 and nothing more.

> I'm not opposed to API changes in general, but I would want people who worked on the dom portion of the package to be pulled in before anything major changes.

The API isn't really changing, it's more so being added to, and adding these changes shouldn't cause a major change, it's a minor change because no publicly exposed function is changing?

Confidenceman02:
@rtfeldman Any chance we could get some direction here? Would be incredibly sad for this to go stale.

rtfeldman:
Sorry, had a lot of other stuff on my plate recently - I'll circle back to this over the weekend!

rtfeldman:
> if we leave out `exactClassName`, we essentially lose the ability to assert the entire class value i.e. "class1 class2 class3 class4" but only for `svg`

🤔 What if we changed it to pass if either the attribute or the property is present and an exact match?

I can think of a downside to having them separate: getting confused when your query says it found no matches even though the property (or attribute) is clearly set, the problem is just that you were querying for the other one...but I'm having trouble thinking of what bugs would be caught by having them separate.

All the scenarios I can come up with seem very unrealistic - e.g. someone sets both `class` and `className` on the same element, but they set them both to different strings, and then use the wrong type of query, which happens to cause them to look more closely at the attributes and realize they've mistakenly set both. It's hard for me to think of something that narrow as being a significant downside.

Maybe there's a scenario I'm not thinking of, though!

Janiczek:
@dinopascale-prima The package has changed maintainers a few times over the past few years. I'll try and take a look at merging this sometime soon, thanks for the ping!

PR #177, opened 2022-03-22
remove the custom test runner and use elm-test

harrysarson:
Deletes a lot of code


harrysarson:
@lydell is this the correct way to install elm-test? I forget if I shoukld use elm-tooling or not

lydell:
Yes, that’s the correct way!

PR #178, opened 2022-03-30
Simplify the test suite machinery

Janiczek:
Now with 100% less Haskell!

Note this clashes with #177.

Screenshot 2022-03-31 at 0 13 43


avh4:
As the original author of the script that's being removed, this looks good to me 👍

tests/run-tests.sh
```diff
@@ -0,0 +1,49 @@
...
+echo -en "${DIM}";
+echo "Running the test suite"
+echo "----------------------------------------------------"
+echo -en "${COLOR_OFF}";
+node elm.js
+
+echo -en "${DIM}";
+echo "----------------------------------------------------"
+echo "Navigating back to the original directory"
+cd - >/dev/null
```
harrysarson, line 48: Why is this needed? The changes to the working dir from within the script won't affect the environment

tests/run-tests.sh
```diff
@@ -0,0 +1,49 @@
...
+DIM="\e[2m";
+COLOR_OFF="\e[0m";
+
+echo -en "${DIM}";
+echo "Navigating to the tests/ directory"
+cd "${0%/*}" # Change current working directory to this one.
+
+LIBRARY_PATH=$(cd .. && pwd)
+PACKAGE_NAME="elm-explorations/test"
+PACKAGE_VERSION="1.2.2"
```
harrysarson, line 12: Can we read this from elm.json with node? I worry we will forget to update this value.

harrysarson (commented):
Looks good, just a couple of minor things

tests/run-tests.sh
Janiczek: I just tried this and you're right! I'll remove that then

tests/run-tests.sh
Janiczek: I'll read this with grep and cut.

harrysarson: Is okay with me! I only suggested node because we already use it below for running elm

harrysarson (approved):

harrysarson:
Feel free to merge, or let me know if I should

PR #182, opened 2022-05-05
Stop crashing in Fuzz.examples

Janiczek:
This was an oversight: crashing is fine inside REPL but there's nothing stopping people from using it in normal Elm applications. Defaulting to an empty list will be safer.


harrysarson (approved):
Good catch

PR #183, opened 2022-06-21
Slight improvement to examples

gampleman:
This makes it so that `Fuzz.examples` produces more variety of example, since it can use the first argument as the seed.

### DIFF
--- a/src/Fuzz.elm
+++ b/src/Fuzz.elm
@@ -1650,7 +1650,7 @@ Useful in REPL:
     [27,25,21,24,29,40,31,23,34,28,28,47,22,36,35,38,33,32,34,29]
         : List Int
 
-Uses seed 0.
+Uses the first argument as the seed as well as the count of examples to generate.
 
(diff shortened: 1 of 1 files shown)


Janiczek:
Thanks, a nice idea - something between the original and "full two degrees of freedom" - having a separate parameter for the seed.

Do you think it warrants updating all the `examples` REPL runs in other functions' doc comments? All the things like
```
    > import Fuzz
    > Fuzz.examples 20 (Fuzz.intRange 20 50)
    [27,25,21,24,29,40,31,23,34,28,28,47,22,36,35,38,33,32,34,29]
        : List Int
```
that are now going to return different data.

Janiczek:
Sorry 🙈

Janiczek:
Actually I grepped the codebase and I thought there will be a bunch, but there's none apart from this `examples` function itself?

Janiczek (approved):
:+1: Thanks @gampleman !

PR #187, opened 2022-07-27
Fuzzer classification and coverage

### LINKED ISSUE #94: Discuss adding labels for fuzzed values
There's a good discussion happening on [discourse](https://discourse.elm-lang.org/t/idea-add-statistical-labeling-to-elm-test) about adding labels and coverage requirements to elm-test. Don't do what I did, watch the 50 minute talk linked to in the first post before trying to say something intelligent.

There may be breaking changes that come out of the discussion to I want to hold open the opportunity to make them.

Janiczek:
Fixes #94
 
- [x] docs
- [x] statistical test, to reduce flakiness
- [x] `Fuzz.labelExamples : Int -> List ( String, a -> Bool ) -> Fuzzer a -> List ( List String, Maybe a )`

---

- [x] `Test.FuzzOptions` now contains `coverage : Coverage a`

---
- [x] `Test.Coverage a = Test.noCoverage | Test.reportCoverage labels | Test.expectCoverage labels`
- [x] `Test.noCoverage : Coverage a`
- [x] `Test.reportCoverage : List ( String, a -> Bool ) -> Coverage a`
  - [x] report after running the test (`Runner.String`)
  - [x] [coordinate with `elm-test` and `elm-test-rs` CLIs](https://discord.com/channels/534524278847045633/762656477608542228/1010642356598493324)
    - [x] write a PR for elm-test-rs's [elm-test-runner](https://github.com/mpizenberg/elm-test-runner/blob/13202e11acb22da5b6914c3b78b5870059f3df3a/src/ElmTestRunner/Result.elm#L78-L80) - https://github.com/mpizenberg/elm-test-runner/pull/13 and https://github.com/mpizenberg/elm-test-rs/pull/118
    - [x] write a PR for [node-test-runner](https://github.com/rtfeldman/node-test-runner/blob/026581eb144a32defd6dae3c3de5c46e5e987527/elm/src/Test/Reporter/TestResults.elm#L59) - https://github.com/rtfeldman/node-test-runner/pull/614
  - [x] off by one error (total number of coverage ticks for `{runs = 100}` is 99)
- [x] `Test.expectCoverage : List ( ExpectedCoverage, String, a -> Bool ) -> Coverage a`
  - [x] silent if coverage is satisfied
  - [x] fail the test if coverage is not satisfied

---

- [x] `Test.ExpectedCoverage = Test.zero | Test.atLeast n | Test.moreThan n`
- [x] `Test.zero : ExpectedCoverage`
  - [x] fail if 1+ values fell into the given category
- [x] `Test.atLeast : Float -> ExpectedCoverage`
  - [x] fail if < the expected number of values fell into the given category
- [x] `Test.moreThanZero : ExpectedCoverage`
  - [x] fail if 0 values fell into the given category

# `Fuzz.labelExamples`

Screenshot 2022-07-29 at 22 38 57

# `Test.reportCoverage`

Screenshot 2022-07-29 at 21 19 13

# `Test.expectCoverage`

## zero
Screenshot 2022-08-20 at 17 25 12

## more than zero
Screenshot 2022-08-20 at 17 25 36

## at least
Screenshot 2022-08-20 at 17 24 44


src/Fuzz.elm
```diff
@@ -773,288 +781,315 @@ helper when creating more complicated fuzzers.
...
-            case fuzzerA prng of
-                Generated a ->
-                    case fuzzerB a.prng of
-                        Generated b ->
-                            Generated
-                                { value = fn a.value b.value
-                                , prng = b.prng
-                                }
+    Fuzzer
+        { coverage = NoCoverage
```
harrysarson, line 820: Is there a reason for `NoCoverage` here?

Janiczek: I don't think there's a way to preserve existing coverage if the types change.

Janiczek: As a simpler example, look at `map` and imagine doing something like
```elm
intFuzzer : Fuzzer Int
intFuzzer =
  Fuzz.intRange 0 9
    |> Fuzz.reportCoverage
      [ ( "below 5", \n -> n < 5 )
      , ( "5 or above", \n -> n >= 5 )
      ]

boolFuzzer : Fuzzer Bool
boolFuzzer =
  intFuzzer
    |> Fuzz.map isEven 
```
Should the user get coverage information about integers and `below 5` / `5 or above` when the end values they are working with are booleans? 🤷

Janiczek: Due to this issue I've moved the coverage API from Fuzz to Test (`FuzzOptions = { runs : Int, coverage : Coverage }`)

Janiczek:
@gampleman raises a good question about using `Expect.classify` or `Expect.cover` instead of having this in `FuzzOptions` (thus living in `Test`). Still unsure about it. I'll first implement the statistical test for reducing flakiness which we need no matter where this lives 🙃

src/Float.elm
```diff
@@ -1,70 +0,0 @@
-module Float exposing (epsilon, infinity, nan, minAbsNormal, minAbsValue, maxAbsValue)
```
Janiczek, line 1: This file is not used anymore (unsure how long that has been the case). Let me know if you want this to live in some other PR.

src/Fuzz.elm
harrysarson: yeah this is a really nice improvement, I like it!

src/Float.elm
harrysarson: bin it!

src/AsciiTable.elm
```diff
@@ -0,0 +1,69 @@
...
+type alias Column a =
+    { toString : a -> String
+    , align : Align
+    }
+
+
+{-| This would normally return just the final String, but we need to postprocess
+the table rows a little (insert a divider row in between).
+-}
+view : List (Column a) -> List a -> List ( a, String )
```
harrysarson, line 18: This return type could be more self explanitory maybe? `List ( a, String )` doesn't tell me much.

It looks like we return this type so that we can split the base reports from the combined one. If so maybe the return type should be:

```
{ baseRows: List String
, combinationRows: List String
}
```

src/Test.elm
```diff
@@ -17,12 +18,15 @@ module Test exposing
...
 @docs fuzz, fuzz2, fuzz3, fuzzWith, FuzzOptions
+@docs Coverage, noCoverage, reportCoverage, expectCoverage
 
 -}
 
 import Expect exposing (Expectation)
 import Fuzz exposing (Fuzzer)
 import Set
+import Test.Coverage exposing (ExpectedCoverage)
+import Test.Coverage.Internal exposing (Coverage)
```
harrysarson, line 29: There is a type alias `Coverage` in this file. We should delete this exposing I think.

src/Test.elm
```diff
@@ -435,6 +467,91 @@ fuzz3 fuzzA fuzzB fuzzC desc =
 
 
 
+-- COVERAGE --
+
+
+{-| With `Coverage` you can observe statistics about your fuzz test inputs and
+assert that a given proportion of test cases belong to a given class.
```
harrysarson, line 474: should we say in this prose what the `a` is?

src/Test/Coverage/Internal.elm
```diff
@@ -0,0 +1,333 @@
...
+> If you are using 'checkCoverage' as part of a test suite, you should
+> be careful not to set @certainty@ too low. If you want, say, a 1% chance
+> of a false positive during a project's lifetime, then @certainty@ should
+> be set to at least @100 \* m \* n@, where @m@ is the number of uses of
+> 'cover' in the test suite, and @n@ is the number of times you expect the
+> test suite to be run during the project's lifetime. The default value
+> is chosen to be big enough for most projects.
+
+In the future we might want to make it configurable, as it is a knob to make
+tests finish faster at the expense of larger probability of false positives.
```
harrysarson, line 96: As a hack the test runners can do some regex magic on the js to set these.

src/AsciiTable.elm
Janiczek: 9f257a2

src/Test.elm
Janiczek: :+1: f050558

src/Test.elm
Janiczek: a2ebe39

src/Test/Coverage/Internal.elm
Janiczek: Well, let's ask the runners whether they want to use this knob 😁 
@lydell @mpizenberg thoughts? Do you want to be able to change
```elm
certainty : Int
certainty =
    10 ^ 9

tolerance : Float
tolerance =
    0.9
```
to different values?

Both will be able to make your tests faster, but it seems to me like it's fast enough (doesn't spin the CPU endlessly) as of right now with these values, so IMHO we could keep it. But eg. Haskell QuickCheck lets you set that as part of their equivalent of `FuzzOptions`.

(To repeat the intuition behind it: the closer your _wanted_ coverage is to the _actual_ coverage, say 50% expected vs 50.5% actual, the more work will the test library need to do to make sure it's not a fluke, and so it will run eg. 100k tests instead of 10k tests. But when you expect eg. 50% coverage and the fuzzer has 40% or 60% coverage, it ends very quickly.)

lydell: I have honestly not followed this closely enough so I don’t really know. But we already do some regex on the JS so adding some more for this if needed sounds good to me!

mpizenberg:
I'm not well versed on this subject and can't check it before mid next
week. So if you feel confident about one way of doing it, I won't be
against. If you prefer waiting for a more detailed answer from me, I'll
need some time.

Janiczek:
Pulling the plug on this as we're nearing the release! There's still a week or so left, please shout if you find an issue or have a question or there is something to reconsider.

PR #191, opened 2022-08-05
Mention elm-test-rs when listing test runners

Janiczek:
We're talking about "a list of runners" in Test.Runner doc comment, but README only contains `node-test-runner`. It would be fair to mention `elm-test-rs` too.

https://github.com/elm-explorations/test/blob/f4e60c2521157d59053cc3580f6309b3c4b38e83/src/Test/Runner.elm#L8-L11

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,11 @@ This package lets you define tests ([Test](https://package.elm-lang.org/packages
 
 To execute your tests and see if they pass you need a program that can consume your tests, run them and report the results.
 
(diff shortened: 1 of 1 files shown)


mpizenberg:
That would be nice, and at the same time, I kind of like the relative peace of having much less users and much less issues raised on the repo ^^

Janiczek:
@mpizenberg Do you want this repo to stay silent about `elm-test-rs` then? I mean, that's totally a reasonable thing to want and I understand the reasons :+1: Let me know!

mpizenberg:
Nah I'm fine with it being mentioned :) if it can help people looking for
some of its features.

harrysarson (approved):

PR #201, opened 2022-10-09
Bump to 2.0.0

Janiczek:


drathier:
Reminder: show off the unicode string fuzzer :)

PR #209, opened 2022-11-22
Add textExactly Selector

### LINKED ISSUE #197: Query.has [Selector.text] does not match exact text
Hello,
Lately while writing some tests where I wanted to make sure that the text inside an html element is **exactly** the same as what I want I found something unexpected (at least to me). It seems that using `Query.has` together with `Selector.text` is not powerful enough to make proper tests that should check for exact match of a text inside. Instead what actually happens is it checks whether the content **contains** a text provided in an argument.

Example:
We want to make sure that the numbers inside our `Html msg` are rounded to 2 decimal places. Let's say that `Query.fromHtml` returns a value representing following html.
```html
        <span class="s1">
            <span class="s3">
                11,22
            </span>
        </span>
  ```
  We pass that into `Query.has [ Selector.text "11,22" ]`.  Everything works fine so far but let's say that the logic changes for some reason and the following is returned.
  ```html
        <span class="s1">
            <span class="s3">
                11,222
            </span>
        </span>
  ```
The same test will not fail and we will not even notice the problem until we find about it during runtime. Is there a way to write this test better? Maybe there should something like `Selector.textExact` :thinking: ? I hope I explained my self well, if you need some clarification, let me know.

### PR DESCRIPTION
**Pull request by Janiczek**
fixes #197

### DIFF
--- a/src/Test/Html/Selector.elm
+++ b/src/Test/Html/Selector.elm
@@ -1,6 +1,6 @@
 module Test.Html.Selector exposing
     ( Selector
-    , tag, text, containing, attribute, all
+    , tag, text, textExactly, containing, attribute, all
     , id, class, classes, exactClassName, style, checked, selected, disabled
     )
 
@@ -11,7 +11,7 @@ module Test.Html.Selector exposing
 
 ## General Selectors
 
-@docs tag, text, containing, attribute, all
+@docs tag, text, textExactly, containing, attribute, all
 
 
 ## Attributes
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by Janiczek** [comment]
@khrynczenko I'll try to ping other maintainers to give it an eye but other than that I'll try to merge it and make a release ~ on  Wednesday.

**Comment by lydell** [comment]
I’m not a native English speaker, but is the -ly in `textExactly` correct? Is `textExact` better? 🤔 🚴

**Comment by Janiczek** [comment]
My thinking wasalong the lines of "Select items where the text is exactly 'foo'".

**Comment by lydell** [comment]
My brain instead went “this is an exact text matcher” 🤷

**Comment by Janiczek** [comment]
IMHO let's do either `exactText` or `textExactly` but not `textExact`. But I know eg. Lue wouldn't agree with me 😁 
WDYT @lydell?

**Review by lydell — APPROVED** [review]
I don’t have strong opinions about the name. Code LGTM (though I’ve never looked at these parts before). The only potential feedback was about -ly.

File: src/Test/Html/Selector.elm [hunk 1]
```diff
@@ -249,13 +249,36 @@ style key value =
...
     Internal.Text
 
 
+{-| Matches elements that have a
+[`text`](http://package.elm-lang.org/packages/elm-lang/html/latest/Html-Attributes#text)
+attribute with _exactly_ the given value (sans leading/trailing whitespace).
+
+`Selector.textExactly "11,22"` will _not_ match `Html.text "11,222"`.
+
+Note this selector _doesn't_ trim the strings:
```
harrysarson, line 270:
```suggestion
Note this selector is whitespace sensitive (it _doesn't_ trim strings prior to checking them):
```

harrysarson (approved):
Think I prefer `exactText`, LGTM though

PR #220, opened 2023-08-17
Implement Fuzz.filterMap

mbaechler:
I needed this function on a project I'm working on so I figured it could be interesting to contribute it back upstream.

edit: By the way, I couldn't run the test locally, I expect to CI to help me about that.

### DIFF
--- a/src/Fuzz.elm
+++ b/src/Fuzz.elm
@@ -9,11 +9,10 @@ module Fuzz exposing
     , array, maybe, result
     , bool, unit, order, weightedBool
     , oneOf, oneOfValues, frequency, frequencyValues
-    , constant, invalid, filter
+    , constant, invalid, filter, filterMap
     , map, map2, map3, map4, map5, map6, map7, map8, andMap
     , andThen, lazy, sequence, traverse
     , fromGenerator
-    , filterMap
     )
 
 {-| This is a library of _fuzzers_ you can use to supply values to your fuzz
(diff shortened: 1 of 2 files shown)


Janiczek:
Hey there! Thanks for the contribution.

Can you please provide more info about your usecase? Just so that I can get my head around when this code would be useful.

Re tests, you can try `./tests/run-tests.sh`. The setup is a bit special as this is elm-explorations package with kernel JS modules.

The code itself looks good, I'd just
- [ ] add the new function to the docs: on line 65, 
```suggestion
@docs constant, invalid, filter, filterMap
```
- [ ] use a more realistic example in the doc for the `filterMap` function. Right now it's just the `filter` doc with `Just a` instead of `True` and `Nothing` instead of `False`. I don't yet know what example that should be though.
- [ ] Add a test for the relationship between filter and filterMap: something along the lines of
```elm
fuzzer |> filterMap f |> map ((/=) Nothing)
===
fuzzer |> filter (f >> (/=) Nothing)
```

lue-bird:
Examples of cases when you want `filterMap` are the same as with `filter`, only that you get a more narrow type when it succeeds, like
```elm
module UnicodeNonLetter exposing (UnicodeNonLetter, fromChar, fuzz)

type UnicodeNonLetter = UnicodeNonLetter Char

fromChar : Char -> Maybe UnicodeNonLetter
fromChar c =
    if (c |> Unicode.isLower |> not) && (c |> Unicode.isUpper |> not) then
        UnicodeNonLetter |> Just

    else
        Nothing


fuzz : Fuzzer UnicodeNonLetter
fuzz =
    Fuzz.char |> Fuzz.filterMap fromChar
```

Janiczek:
Sounds lovely - that would be a nice example to have in the docs :+1:

mbaechler:
Hi @Janiczek ,

I finally found some time to update my PR, it should be good now.

Let me know if there are still things to fix.

Janiczek:
Hey @mbaechler, the CI tests failed because of a forgotten stale version in tests/elm.json. I've fixed it here: https://github.com/elm-explorations/test/commit/81b15e2e9ae39a3a437e1dc80ede11acd86453da

I'm gonna pull master into your branch and try the tests again.

Janiczek:
Thanks for your work on this!

PR #233, opened 2024-02-21
Add missing imports in documentation

jfmengels:
I've noticed a few missing imports in the examples.

A few notes: The imports are not sorted the way that `elm-format` sorts them. Would you like to see this be done as well?

I've only done it for this module, but I've seen some examples in `Test.Html.Query` where imports are missing and/or lead to confusion and invalid code. (See for instance `Test.Html.Query.keep` where `text` is used from both `Html` and `Test.Html.Selector`, and similarly `Test.Html.Query.children` uses `class` in the same invalid way). Let me know if you want me to solve this (and if so, indicate how you would prefer to see these collisions resolved). Maybe other modules have import issues, I haven't looked at it yet.

### DIFF
--- a/src/Test.elm
+++ b/src/Test.elm
@@ -141,7 +141,7 @@ describe untrimmedDesc tests =
 {-| Return a [`Test`](#Test) that evaluates a single
 [`Expectation`](../Expect#Expectation).
 
-    import Test exposing (fuzz)
+    import Test exposing (test)
     import Expect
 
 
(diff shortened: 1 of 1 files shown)


lue-bird (approved):

Janiczek:
Thanks Jeroen!

> The imports are not sorted the way that elm-format sorts them. Would you like to see this be done as well?

Yeah that would be nice!

Janiczek:
Let's keep Test.Html.Query for some other day :+1:

PR #243, opened 2025-04-29
Support providing a fake Browser.Navigation.Key in tests

### LINKED ISSUE #24: Anything with Browser.Navigation.Key cannot be tested
When using Browser.application I will put the Key in my model for later use.

When testing I want to create a model, but I don't have a way to get a Key. 
So I cannot test functions that take my Model.

Can test provide a way to get a navigation key?

Originally opened this here https://github.com/elm/browser/issues/30

jfmengels:
Closes #24

Adds `Test.withNavigationKey` which provide a [`Browser.Navigation.Key`](https://package.elm-lang.org/packages/elm/browser/latest/Browser-Navigation#Key) to tests.

This functions is designed such that the key can't be used in production code in a way that will have a noticeable effect (one could generate a test in production code, but they can't really do anything with a test).

Without this, the workaround is to define a wrapper type
```elm
type Key
  = RealKey Browser.Navigation.Key
  | TestKey
```
as well as wrapper functions for `Browser.Navigation` functions (such as `pushUrl`) that accept this type, where they'll do nothing if the key is not a real key.

---

Implemention-wise, the key is a simple function with a side-effect and no input ([source](https://github.com/elm/browser/blob/1.0.2/src/Elm/Kernel/Browser.js#L146)).
I've introduced a new test key which is a plain noop function, which should be equivalent in all "noticeable" ways (even through `Debug.log`).

**Releasing this functionality will lock the implementation of `elm/browser`'s key to such a function**, at least until it releases a new major version.

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -17,6 +17,7 @@
     ],
     "elm-version": "0.19.0 <= v < 0.20.0",
     "dependencies": {
+        "elm/browser": "1.0.1 <= v < 2.0.0",
         "elm/bytes": "1.0.8 <= v < 2.0.0",
         "elm/core": "1.0.0 <= v < 2.0.0",
         "elm/html": "1.0.0 <= v < 2.0.0",
--- a/src/Elm/Kernel/Test.js
+++ b/src/Elm/Kernel/Test.js
@@ -16,3 +16,5 @@ function _Test_runThunk(thunk)
     return __Result_Err(err.toString());
   }
 }
+
+function _Test_navigationKey() {}
--- a/src/Test.elm
+++ b/src/Test.elm
@@ -1,5 +1,6 @@
 module Test exposing
     ( Test, test
+    , withNavigationKey
     , describe, concat, todo, skip, only
     , fuzz, fuzz2, fuzz3, fuzzWith, FuzzOptions
     , Distribution, noDistribution, reportDistribution, expectDistribution
@@ -8,6 +9,7 @@ module Test exposing
 {-| A module containing functions for creating and managing tests.
 
 @docs Test, test
+@docs withNavigationKey
 
 
 ## Organizing Tests
@@ -22,6 +24,8 @@ module Test exposing
 
 -}
 
+import Browser.Navigation
+import Elm.Kernel.Test
 import Expect exposing (Expectation)
 import Fuzz exposing (Fuzzer)
 import Set
@@ -162,6 +166,34 @@ test untrimmedDesc thunk =
 
     else
         Internal.ElmTestVariant__Labeled desc (Internal.ElmTestVariant__UnitTest (\() -> thunk ()))
+
+
+{-| Provide a [`Browser.Navigation.Key`](https://package.elm-lang.org/packages/elm/browser/latest/Browser-Navigation#Key)
+value to your tests.
+
+You might need such a key if you're testing functions where a key is contained in the arguments, for instance when using
+[`Browser.application`](https://package.elm-lang.org/packages/elm/browser/latest/Browser#application)
+and storing the key in the `Model`.
+
+    import Expect
+    import Test exposing (test)
+
+    Test.withNavigationKey <|
+        \key ->
+            test "functionToTest returns 0 when given 0" <|
+                \() ->
+                    functionToTest key 0
+                        |> Expect.equal 0
+
+-}
+withNavigationKey : (Browser.Navigation.Key -> Test) -> Test
+withNavigationKey testFn =
+    testFn testNavigationKey
+
+
+testNavigationKey : Browser.Navigation.Key
+testNavigationKey =
+    Elm.Kernel.Test.navigationKey
 
 
 {-| Returns a [`Test`](#Test) that is "TODO" (not yet implemented). These tests
--- a/tests/src/BrowserTests.elm
+++ b/tests/src/BrowserTests.elm
@@ -1,0 +1,54 @@
+module BrowserTests exposing (all)
+
+import Browser.Navigation
+import Expect
+import Fuzz
+import Test exposing (Test, describe)
+
+
+all : Test
+all =
+    Test.withNavigationKey <|
+        \navKey ->
+            Test.describe "Test.withNavigationKey"
+                [ regularTest navKey
+                , fuzzTest navKey
+                ]
+
+
+regularTest : Browser.Navigation.Key -> Test
+regularTest navKey =
+    Test.test "(regular test) storing, using or comparing the key should not have any noticeable effect" <|
+        \() ->
+            let
+                init : { navKey : Browser.Navigation.Key }
+                init =
+                    { navKey = navKey }
+            in
+            ( init, cmdsUsingKey navKey )
+                |> Tuple.first
+                |> Expect.equal { navKey = navKey }
+
+
+fuzzTest : Browser.Navigation.Key -> Test
+fuzzTest navKey =
+    Test.fuzz Fuzz.int "(fuzz) storing, using or comparing the key should not have any noticeable effect" <|
+        \number ->
+            let
+                init : { navKey : Browser.Navigation.Key, number : Int }
+                init =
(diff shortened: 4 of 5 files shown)


elm.json
```diff
@@ -17,6 +17,7 @@
     ],
     "elm-version": "0.19.0 <= v < 0.20.0",
     "dependencies": {
+        "elm/browser": "1.0.2 <= v < 2.0.0",
```
Janiczek, line 20: I wonder whether this is as low as we can go, or whether 1.0.1 or 1.0.2 have something important we want to encourage adoption of.

src/Test.elm
```diff
@@ -164,6 +170,40 @@ test untrimmedDesc thunk =
         Internal.ElmTestVariant__Labeled desc (Internal.ElmTestVariant__UnitTest (\() -> [ thunk () ]))
 
 
+{-| Return a [`Test`](#Test) that requires a [`Browser.Navigation.Key`](https://package.elm-lang.org/packages/elm/browser/latest/Browser-Navigation#Key)
```
Janiczek, line 173: "requires" was surprising to me. I'd say it _provides_ a key instead?

I'm wondering whether it would be benefitial to also mention elm-program-test or Browser.application `init` functions. They're not the only scenario an user might need the Key for, but might be the most common one?

Janiczek: A third note: Is there a gap in the API now where `Test.testWith` users still can't get to the key? This might be getting into a "we need to redesign `testWith` to be a builder pattern instead" territory, IDK yet... 🤔

jfmengels: I've improved the wording a bit, let me know what you think. (Don't know why I used "require" :man_shrugging: )

I've mentioned Browser.application, but I've purposefully ignored `elm-program-test` because it is not a core library. Someone could fork that project or make something similar but better one day and this package would push towards one solution rather than another, which I don't think is fair.

And yes, I think there's a gap in the API. It definitely reeks of multiple combinations of options, so a builder-like pattern is starting to feel like it could be justified. I don't think we're there yet but your mileage may vary. If there was one more kind of option then I definitely think there would have to be some re-design.

So should we add a new function? and name it... `testWithWithKey`? :nauseated_face:

Janiczek: If a builder pattern is possible, I'd definitely prefer it to `testWith` (where you suddenly have to care about distributions even if you just want to up the run count).

If we can do that without a final `|> build`, that would be (roughly) 100000x better.

I can try and brainstorm something on another branch.

jfmengels: Just to clarify, when you said `Test.testWith`, did you mean `Test.fuzzWith`?

src/Elm/Kernel/Test.js
```diff
@@ -16,3 +16,5 @@ function _Test_runThunk(thunk)
     return __Result_Err(err.toString());
   }
 }
+
+function _Test_navigationKey() {}
```
lydell, line 20: Here’s an idea I just got:

```suggestion
function _Test_navigationKey() {
  throw new Error('This is a test Browser.Navigation.Key that must not be used for real!');
}
```

And we could expose it like so:

```elm
-- module Test

{-| Exposed to users of the package! -}
navigationKey : Browser.Navigation.Key
navigationKey =
		Elm.Kernel.Test.navigationKey
```

Then there is no need for new `testWithKey` and `fuzzWithKey` functions.

If someone were to try to use it for real in an app, it would _compile._ But at runtime, it would run a `history` method and then throw an error, breaking the app. (We should verify that it breaks the app enough, though.)

https://github.com/elm/browser/blob/1.0.2/src/Elm/Kernel/Browser.js#L190-L212

Alternatively, instead of throwing an error we could do something drastic like replacing the whole `document.body` with an error message.

Not sure if this is good or not, but it _is_ tempting.

jfmengels: Interesting idea!

Maybe we could make it fail (throw) unless something was set by the test runner (ex: `globalThis.ELM_TESTING === true`), though all test runners would have to change something which is painful unless there's already something we can check for.

lydell: Oh, is it possible to execute the key function in tests? Or why would we need a guard?

jfmengels: I am not sure what you mean. The guard is meant is to prevent people using this value in application (non-test) code.

Alternatively, we could have a design like this, resembling the original `testWithKey` of this PR, but with a value that allows you to build any "dangerous" value.

```elm
-- Names to be bikeshedded
testWithData : String -> (MasterKey -> Expectation) -> Test

test "title" <|
  \masterKey ->
    let navigationKey = Test.navigationKey masterKey
    in
    -- ...
```

That way, if new values like `Navigation.Key` come up later in Elm's public API, we won't have to add new variants of `test`, we will only have to introduce a new function:

```elm
-- Test.elm
newResource : MasterKey -> Resource
```

But it does mean we will need to solve the problem of all the `test`/`fuzz` variants that blocked this original PR.

---

People can easily recreate `testWithKey` with this API as well:

```elm
testWithKey title fn =
  Test.testWithData title (Test.navigationKey >> fn)
```

lydell: > The guard is meant is to prevent people using this value in application (non-test) code.

How? I didn’t understand what `globalThis.ELM_TESTING === true` is supposed to do. (I made a guess, but just confused you instead.)

jfmengels: The idea was to do something like
```elm
function createNavigationKey() {
  if (globalThis.ELM_TESTING !== true) {
    throw new Error("This is only meant to be used in test code!");
  }
  return dummyNavigationKey;
}
```
and have the test runners define this global constant somewhere so that this check would pass: `globalThis.ELM_TESTING = true;`.

That said, with the proposal in my previous comment, this is unnecessary.

lydell: Here’s one more idea for how to provide the key without making new `test` and `fuzz` functions:

```elm
tests =
	withKey <| \navKey ->
        describe "my tests"
            [
                -- tests…
            ]
```

jfmengels: Yeah... that would work quite well!

```elm
withKey : (Navigation.Key -> Test) -> Test
```

What do you think @Janiczek?

jfmengels:
Alright, I've simplified this PR to just adding `Test.withNavigationKey`. It's a lot simpler and flexible.

Previous attempt can (for now) be found at https://github.com/elm-explorations/test/compare/master...jfmengels:test:navigation-key-old

src/Test.elm
```diff
@@ -164,6 +168,34 @@ test untrimmedDesc thunk =
...
+[`Browser.application`](https://package.elm-lang.org/packages/elm/browser/latest/Browser#application)
+and storing the key in the `Model`.
+
+    import Expect
+    import Test exposing (test)
+
+    Test.withNavigationKey <|
+        \key ->
+            test "functionToTest returns 0 when given 0" <|
+                \key ->
```
lydell:
```suggestion
                \() ->
```

elm.json
lydell: It looks like `Browser.Navigation.Key` used to be an object, not a function:

https://github.com/elm/browser/commit/544138a79d3e80b7ff8ccfa11e2a6f62eec1a87c

That commit is dated Jul 30, 2018 and 1.0.1 is the first tag after that: https://github.com/elm/browser/tags

So we could _potentially_ lower this to 1.0.1 (after double-checking that stuff actually works), but I don't think there is any gain doing this in practice, since 1.0.2 was released in late 2019.

jfmengels: I've lowered it to 1.0.1, that seems to work fine. In practice I believe we could lower it to 1.0.0 without any noticeable effect, except that `Debug.log` would show something slightly different just for `1.0.0`. I think 1.0.1 is fine but happy to change it.

lydell:
This probably closes https://github.com/elm-explorations/test/issues/24

PR #244, opened 2025-05-01
Simplify expectations to a single expectation

jfmengels:
Internally, the tests were expecting a list of `Expectation`, but it seems it's always a singleton list.
We can therefore remove the surrounding list, which simplifies the internals a bit.


Janiczek:
Thank you @jfmengels! Was this found with elm-review? 😄

jfmengels:
Unfortunately this was a manual find 😄

Issue #246, opened 2025-07-12
`Selector.class` only matches `className` property, not `class` attribute


lydell:
You know how you can use `Html.Attributes.class` multiple times to stack them? Like `[ Html.Attributes.class "one", Html.Attribute.class "two" ]` results in `class="one two"`. That’s actually not an elm/html feature, it’s an elm/virtual-dom feature. It’s implemented for both [the `className` property](https://github.com/elm/virtual-dom/blob/1.0.4/src/Elm/Kernel/VirtualDom.js#L399-L400) and the [the `class` attribute](https://github.com/elm/virtual-dom/blob/1.0.4/src/Elm/Kernel/VirtualDom.js#L407-L408), even though [Html.Attributes.class](https://github.com/elm/html/blob/1.0.0/src/Html/Attributes.elm#L205-L207) is currently implemented only using the `className` property.

In other words, `VirtualDom.property "className" (Json.Encode.string "my-class")` and `VirtualDom.attribute "class" "my-class"` are equal citizens. Except in elm-explorations/test. `Query.has [ Selector.class "my-class" ]` only works if the class was set using `className`. Which it is 99.99 % of the time since `Html.Attributes.class` is implemented using `className`.

So to follow elm/virtual-dom’s lead, elm-explorations/test should do the same.

I also changed `Html.Attributes.class` to use the `class` attribute in https://github.com/elm/html/pull/259. A user of my forked packages let me know that their tests stopped passing when they started using my forks. That led me to implement the following workaround to stay backwards compatible:

https://github.com/lydell/virtual-dom/commit/81c070c9edc4e7bf13d5ded8dee1d1dc69b26877

(That link also explains why I switched from `className` to `class`.) So that’s also a reason to do this.

PR #247, opened 2025-07-12
Support classes set using the `class` attribute

lydell:
Closes https://github.com/elm-explorations/test/issues/246


Janiczek:
Thanks for your work on this @lydell, this is now live in 2.2.1.

PR #249, opened 2025-08-01
Add golden file tests

micahhahn:
This PR adds a new primitive equalToFile to the Expect module. It also adds a prettyPrintSingle method

This primitive allows us to write golden file / snapshot tests in Elm. These are particularly useful at capturing the current outputs of a program and then failing if the outputs change.

NoRedInk in particular has an interest in snapshotting the HTML at various points in elm-tests to a file. I've also added a prettyPrintSingle function to the query module which exposes the internal query pretty print function. I think this will be useful generally for testing to be able to capture focus of a query in a complicated test (especially with elm-program-test). NRI would also see value in dumping the entire HTML of a page to a golden file so that we can do visual diffs of our pages in their various states.

Note that the code its current state is very rough - the kernel code added presumes that the tests are being run in the context of a node environment. That's always true for [node-test-runner](https://github.com/rtfeldman/node-test-runner) but AFAIK not always true for [elm-test-rs](https://github.com/mpizenberg/elm-test-rs). I'm very open to any ideas that would make this more robust.

I'm also aware that doing the IO direclty inside of the elm-explorations/test probably prevents the test runners from doing proper watching on these golden files. Definitely open to ideas on approaches that would let the actual file IO be handled by the runner itself.

### DIFF
--- a/src/Elm/Kernel/Test.js
+++ b/src/Elm/Kernel/Test.js
@@ -5,6 +5,7 @@ import Result exposing (Err, Ok)
 
 */
 
+const fs = require('fs');
 
 function _Test_runThunk(thunk)
 {
@@ -16,3 +17,26 @@ function _Test_runThunk(thunk)
     return __Result_Err(err.toString());
   }
 }
+
+function _Test_readFile(filePath)
+{
+    try {
+        return __Result_Ok(fs.readFileSync(filePath, { encoding: 'utf8' }));
+    }
+    catch (err)
+    {
+        return __Result_Err(err.toString())
+    }
+}
+
+var _Test_writeFile = F2(function(filePath, contents)
+{
+    try {
+        fs.writeFileSync(filePath, contents);
+        return __Result_Ok(__Utils_Tuple0);
+    }
+    catch (err)
+    {
+        return __Result_Err(err.toString());
+    }
+})
--- a/src/Elm/Kernel/Test.js
+++ b/src/Elm/Kernel/Test.js
@@ -106,6 +106,36 @@ var _Test_writeFile = F2(function(filePath, contents)
     return WriteFile(path.resolve("tests"), filePath, contents);
 })
 
+function _Test_deleteFile(filePath)
+{
+    // Test for this early as `resolve` will strip training slashes
+    if (filePath.slice(-1) == path.sep) {
+        return __Result_Err(__File_IsDirectory);
+    }
+
+    // Protect against deleting files above the "tests" directory
+    const testsPath = path.resolve("tests");
+    const fullPath = path.resolve(testsPath, filePath);
+
+    if (!fullPath.startsWith(testsPath))
+    {
+        return __Result_Err(__File_PathEscapesDirectory);
+    }
+
+    try {
+        fs.unlinkSync(fullPath);
+    }
+    catch (err)
+    {
+        if (err.code == "ENOENT"){
+            return __Result_Err(__File_FileNotFound);
+        }
+        else {
+            return __Result_Err(__File_GeneralFileError(err.toString()));
+        }
+    }
+}
+
 var overwriteGoldenFiles = null;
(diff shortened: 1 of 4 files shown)


Janiczek (commented):
Hey @micahhahn, apologies for the radio silence here on this PR!

I've so far only read through the diff, I'll have to try it out etc.

But some thoughts so far:

1) **Defining a `File` module.** Could the module name `File` cause collision issues for test suites in projects that use `elm/file` as dependency?
    * Perhaps we could make this be `Test.File` or `Test.GoldenFile` or `Test.SnapshotFile` or something similar.

2) **Node/Deno.** When you say this expects Node (fine for `elm-test` but not always for `elm-test-rs`), by this you mean that `elm-test-rs` can be configured to use Deno, right?
    * That could perhaps be solved for, the JS code could check which environment it's in and use different platform APIs.

3) **Browser environments.** For old versions of `elm-explorations/test`, browser test runners like [`jgrenat/elm-html-test-runner`](https://package.elm-lang.org/packages/jgrenat/elm-html-test-runner/latest/Test-Runner-Html) do exist; for the current 2.0.0+ version there's no nice ready-to-use runner, but running tests in the browser is still supported - see [an example Ellie](https://ellie-app.com/xP2HRfmCby6a1).
    * So, how should `Expect.equalToFile` work in the browser? Should it always fail? Always behave as if the test was `skip`ed? Always pass? Should we try and do some phantom type magic to distinguish between Node / Deno / browser test runners on the type level, and disallow `Expect.equalToFile` in browser environment?

4) **Purpose of the `File` module.** Right now it seems to me it tries to do both generic "read/write files" tasks and specific "deal with `.failed.html` logic" ones.
    * Maybe the renaming from 1) will later give us some clarity on if the contents need to be refactored somehow.

5) **Example usages of `Expect.equalToFile`.** I can imagine how this gets used with raw strings. Can you expand (either in doc comment or here in the PR) on how you imagine this to be used with HTML tests? If you're already using a vendored version of this repo at NRI for snapshot tests, can you show some real-world HTML examples?
    * Maybe we can find a better API (ie. adding `htmlEqualToFile : String -> Html -> Expectation` and making _that_ do `.html` specific logic, and making `equalToFile` instead  do `.txt` logic).
    * Since `Expect.equal` already works with arbitrary `a` values, perhaps we could even aim for `equalToFile : String -> a -> Expectation`...

dillonkearns:
@micahhahn I made a PR to expose some functions that would support running `elm-program-test` cases with `elm-snapshot`: https://github.com/avh4/elm-program-test/pull/191

If that's merged then I imagine that would give you similar functionality to what you're describing.

dillonkearns:
@micahhahn update: I published an API for running snapshot tests against the view and model after an `elm-program-test` run: https://package.elm-lang.org/packages/dillonkearns/elm-snapshot/latest/Snapshot-ProgramTest.

micahhahn:
@dillonkearns 

Finally got a change to look at this... very promising! 

I am hitting a snag trying to test this with our codebase.  We use elm-css quite a bit and some types there are opaque which causes compilation failures generating wire stuff with lamdera.  I've confirmed that we can compile with `--no-wire` passed to `lamdera make`.  

Does `elm-pages run` actually *need* the Lamdera wire types? If not could you add an option to `elm-pages run` to disable their generation so that I can give this a proper test? 

Thanks!

dillonkearns:
Hey @micahhahn! Interesting, I dug into that and I believe it is a lamdera compiler bug which I found a fix for, PR is here: https://github.com/lamdera/compiler/pull/85. I don't want to add to much noise to this thread, so if there's anything else to discuss, could you open an issue in the elm-snapshot repo to discuss further (happy to discuss on Slack or Discord as well). Would love to hear your feedback and help make sure you are not blocked from using elm-snapshot! Keep me posted.

PR #250, opened 2025-08-16
Use _VirtualDom_toTest instead of decoding hacks

lydell:
Requires this branch of virtual-dom: https://github.com/lydell/virtual-dom/tree/tostring

Now we don’t depend on internals in elm/virtual-dom that we shouldn’t depend on anymore.

This is the dream – I wanted to see what it would look like.


Janiczek:
I might have asked somewhere already - is there some plan between you and @evancz to merge `lydell/virtual-dom` into `elm/virtual-dom`?

I'm asking to learn whether this is the way forward in the future ("we're planning to merge virtual-dom") or purely a "what-if" scenario.

lydell:
This is purely a what-if scenario. My branch contains a safer, faster and less buggy `toString` which would benefit elm-pages, as well as a sane way for elm-test to do its thing. Both of which will probably never see the light of day. But it was very fun to program.

PR #258, opened 2026-08-12
Various perf improvements

jfmengels:
I went through the repo and found a bunch of (micro-)optimizations and some cleanup changes. (No AI involved)

I tried to make it so no commit should ever make any behavioral change. If you find some, then that is a mistake on my part that should be fixed. Some commits make the code worse but are improved in the following commits, usually in an attempt to explain through tiny steps that code is equivalent. In some cases, the code ends up maybe harder to read but should be faster.

Feel free to ask me to make changes, and/or to pick or discard whatever you (don't) like.


Janiczek:
Fully merged in the split PRs. Thank you so much @jfmengels !

Issue #97, opened 2019-06-19
Move the name parameter to the first place in all `fuzz` functions


nikita-volkov:
The first thing the user wants to see when he skims through a suite definition is the names of things. However the current implementation gives preference to fuzzers, which hardly can describe what the test is about, leaving the verbal description of the test to hang somewhere in the end of the line or even on a new line.

E.g., here's the way things have to be now:

```elm
fuzz2 (Fuzz.list Fuzz.int) Fuzz.int "List.reverse never influences List.member" <|
  ...
```

Here's what I suggest:
```elm
fuzz2 "List.reverse never influences List.member" (Fuzz.list Fuzz.int) Fuzz.int <|
  ...
```

mgold:
Not a bad idea; I've noticed this before. Applying label for major API change.

harrysarson:
Is this something we want to do? I can work on a patch for it if so.

drathier:
There's a good argument against it in that this will break pretty much every single test suite in existance. However, if we do want to perform this change, the sooner the better.

harrysarson:
I wonder if we could work with elm-analyzer (or a different tool) to automatically update.

Janiczek:
Will be fixed in 446c978 (not part of a release yet). We'll also need to make an elm-review rule to automate the upgrade

PR #259, opened 2026-08-16
Add `repeat` for parametrized tests and fuzz test examples

lydell:
This PR adds a small convenience function called `repeat` (and one called `testWith`, which is `test` but shaped to fit `repeat`). This does not unlock anything that wasn’t already possible in user land – it’s all about convenience and teaching.

Me and @Janiczek were chatting, and Martin mentioned that when a fuzz test failed, he was missing a convenient way of adding that case as a regression test. `repeat` to the rescue!

```elm
    repeat "String.words never returns an empty list"
        [ testWith " \n\t" "only whitespace" -- regression test
        , fuzz Fuzz.string "fuzz"
        ]
    <|
        \string ->
            String.words string
                |> Expect.notEqual []
```

What I really like about this solution (and that I haven’t seen in ad-hoc implementations of this in test suites):

- You can use `skip` and `only`. No commenting out stuff in the list!
- IDE:s have a chance to find a location for each sub test. Ad-hoc implementations typically take a list of data – instead of using test functions in the list, like above – which means that there isn’t an “anchor” for each test that IDE:s can use. Also, by having this in the package, IDE:s have blessed names to look for (in addition to `test`, `fuzz`, `describe` etc.), whereas user land functions don’t really have that possibility.

### DIFF
--- a/src/Test.elm
+++ b/src/Test.elm
@@ -3,7 +3,7 @@ module Test exposing
     , describe, concat, todo, skip, only
     , fuzz, fuzz2, fuzz3, fuzzWith, FuzzOptions
     , Distribution, noDistribution, reportDistribution, expectDistribution
-    , repeat, testWith
+    , repeat, testWithValue
     )
 
 {-| A module containing functions for creating and managing tests.
@@ -24,7 +24,7 @@ module Test exposing
 
 ## Repeating Tests
 
-@docs repeat, testWith
+@docs repeat, testWithValue
 
 -}
 
@@ -572,10 +572,10 @@ This is useful if you want to run the same test with different inputs – also
 known as _parametrized tests:_
 
     repeat "checking if a string is whitespace only"
-        [ testWith "empty string" { input = "", expected = True }
-        , testWith "space, newline, tab" { input = " \n\t", expected = True }
-        , testWith "only non-whitespace" { input = "a", expected = False }
-        , testWith "mixed" { input = "a b", expected = False }
+        [ testWithValue "empty string" { input = "", expected = True }
+        , testWithValue "space, newline, tab" { input = " \n\t", expected = True }
+        , testWithValue "only non-whitespace" { input = "a", expected = False }
+        , testWithValue "mixed" { input = "a b", expected = False }
         ]
     <|
         \{ input, expected } ->
@@ -596,7 +596,7 @@ guarantee that specific input will ever be generated again. In this case, you
 can use `repeat` to save such inputs as explicit regression tests:
 
     repeat "String.words never returns an empty list"
-        [ testWith "only whitespace" " \n\t"
+        [ testWithValue "only whitespace" " \n\t"
         , fuzz Fuzz.string "fuzz"
         ]
     <|
@@ -611,7 +611,7 @@ If you want to [skip](#skip) tests, or focus on tests with [only](#only), use
 function composition (`<<`):
 
     repeat "String.words never returns an empty list"
-        [ only << testWith "only whitespace" " \n\t"
+        [ only << testWithValue "only whitespace" " \n\t"
         , fuzz Fuzz.string "fuzz"
         ]
     <|
@@ -628,8 +628,8 @@ repeat desc tests thunk =
 {-| This is a small wrapper around [test](#test), which is supposed to be used with [repeat](#repeat).
 See that function for examples.
 -}
-testWith : String -> a -> (a -> Expectation) -> Test
-testWith desc a thunk =
+testWithValue : String -> a -> (a -> Expectation) -> Test
+testWithValue desc a thunk =
     test desc (\() -> thunk a)
 
 
--- a/src/Test.elm
+++ b/src/Test.elm
@@ -572,10 +572,10 @@ This is useful if you want to run the same test with different inputs – also
 known as _parametrized tests:_
 
     repeat "checking if a string is whitespace only"
-        [ testWith { input = "", expected = True } "empty string"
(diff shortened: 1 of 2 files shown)


jfmengels:
Very interesting!

One thing I would change is the order of arguments for `testWith`: I would put the title first, that way even if the data is large, the title is at the top like all tests.

```elm
-- Currently
testWith
  """Multi
       line
       piece
       of
       data
  """
  "only whitespace"

-- Suggested
testWith "only whitespace"
  """Multi
       line
       piece
       of
       data
  """
```

lydell:
Yep, I would also like it the other way. But I followed the `fuzz` order. (Which is wrong, too, IMO. The title should always come first.)

Maybe it’s better to do `testWith` “correctly” rather than “consistently”?

jfmengels:
I agree about the arguments for the fuzz function being out of order. That said, the title remains near the top in the usual (non-`repeat`) usage. Maybe that's a possible breaking change to suggest for the next major version?

For `testWith`, I think it would be better the way I suggested.

lydell:
> That said, the title remains near the top in the usual (non-`repeat`) usage.

I’m not quite following here. Do you have an example?

jfmengels:
What I meant is that when defining a fuzz test such as
```elm
fuzz
  myFuzzer
  "title"
  <|
    \data -> ...
```
the title is still "near" the top of the test definition. Not as high as it could be (`fuzz "title" myFuzzer`) though.

lydell:
Btw, is it confusing that `testWith` and `fuzzWith` aren’t really the same? `testWith` is `testWithValue` while `fuzzWith` is `fuzzWithOptions`.

jfmengels:
Right, that's a good point. `testWithValue` could be a better name indeed.

lydell:
> Maybe that's a possible breaking change to suggest for the next major version

Btw, I found an issue about this: https://github.com/elm-explorations/test/issues/97

src/Test.elm
```diff
@@ -557,6 +563,75 @@ expectDistribution =
...
+
+You can also use this for [fuzz](#fuzz) tests. Let’s say your fuzz test found an
+issue after a _long_ time, and you fix it. Do you have confidence it will never
+regress? After all, finding the issue took some amount of luck, and there is no
+guarantee that specific input will ever be generated again. In this case, you
+can use `repeat` to save such inputs as explicit regression tests:
+
+    repeat "String.words never returns an empty list"
+        [ testWithValue "only whitespace" " \n\t"
+        , fuzz "fuzz" Fuzz.string
```
lydell, line 599: @Janiczek  Aren’t these arguments to [fuzz](https://elm.dmy.fr/packages/elm-explorations/test/latest/Test?q=fuzz#fuzz) in the wrong order?

(You changed the order to the one I would _like,_ but not the order that the function actually uses _today._)

Janiczek:
I've made the `fuzz` argument order change that #97 talks about, in 446c978 (on master, not in any release yet, will be a breaking change).

I've made a commit here that fixes your docs to make things read like `fuzz "fuzz" ...`.

---

I'm still mulling this thing over - it's nicely general in the sense that you can combine `test`, `fuzz`, etc. together in the list.

It solves the "let's add a regression test" usecase nicely.

But for parameterized tests, I think another helper could serve them better: something like [Test.Table](https://package.elm-lang.org/packages/drathier/elm-test-tables/latest/Test-Table) or what I have in a few places:

1. [have a list of just data](https://github.com/Janiczek/elm-syntax-type-inference/blob/17075bc66188d5fd93b7a9783dfc561597fe3e50/tests/TypeInferenceTests.elm#L323-L325) (eg. `(testName, input, expectedOutput)` or just `(input, expectedOutput)` where the test name gets generated from the input)

2. [have a function](https://github.com/Janiczek/elm-syntax-type-inference/blob/17075bc66188d5fd93b7a9783dfc561597fe3e50/tests/TypeInferenceTests.elm#L81) going from the "test case" (the tuple in the example list) to a `Test`

3. [combine them together](https://github.com/Janiczek/elm-syntax-type-inference/blob/17075bc66188d5fd93b7a9783dfc561597fe3e50/tests/TypeInferenceTests.elm#L465) in a `Test.describe` and `List.map`

```elm
testCases : List (Int,Int,Int)
testCases = 
  [ (1,1,2)
  , (5,0,5)
  ]

testCase : (Int,Int,Int) -> Test
testCase (a,b,expectedSum) =
  Test.test (Debug.toString (a,b)) <|
    \() ->
      (a+b)
        |> Expect.equal expectedSum

myTests : Test
myTests =
  Test.describe "addition" (List.map testCase testCases)
```

(Ignore my naming...)

I think we could have a blessed helper for parameterized tests like this, and position the `repeat` thing more towards the regression tests. (Or maybe there's another usecase?)

Of course, parameterized tests can be also implemented using `repeat`:

```elm
myTests : Test
myTests =
  Test.repeat "addition" 
    (testCases 
      |> List.map
           (\((a,b,expectedSum) as testCase) -> 
             Test.testWithValue (Debug.toString (a,b)) testCase
           ) 
    )
    <|
      \(a,b,expectedSum) ->
        (a+b)
          |> Expect.equal expectedSum
```

but it's a little bit weird.

Anyways, this PR is probably all good, there's just opportunities for a nicer API for parameterized tests.

src/Test.elm
Janiczek: see 446c978 and my comment below. you were a few seconds faster :D

lydell:
Cool, thanks for reviewing! 👍

PR #200, opened 2022-10-07
add downcast test api to help testrunners

harrysarson:

PR #260, opened 2026-08-16
Add Test.RunnerV2 module

### LINKED ISSUE #154: capture Debug.log and present it with a failing test
When I'm trying to figure out why a test is failing, I'll often put `Debug.log` statements into the things I'm modifying to see what's up. If I run my whole test suite after that, I get way more log messages than I want, with no indication of which ones are actually relevant to the test failure I'm trying to debug.

I'd really like it if, instead, elm-test would capture stdout and associate it with a given test run. A few other test frameworks I've used (py.test maybe? but it's been a while) do this and it's super nice.

If I could have a pony, I'd also like this for fuzz tests. I'd like to only see the debug messages from the final shrunk failure case!

lydell:
This deprecates the `Test.Runner` module and adds `Test.RunnerV2`. This is a backwards-compatible change.

node-test-runner PR: https://github.com/rtfeldman/node-test-runner/pull/686. The idea is to _require_ a recent enough version of elm-explorations/test in node-test-runner (updating the package in projects is a no-brainer), but keep compatibility in this package, so we don’t leave elm-test-rs behind.

Changes in `Test.RunnerV2`:

- Removed seed distribution. All fuzz tests now run with the same seed. This means that moving a test, adding another fuzz test, or commenting out a fuzz test no longer can cause fuzz tests to behave differently, even though you passed a fixed seed. This also got rid of a lot of complexity.
- Unit tests and fuzz tests are now returned separately. This allows a runner to more cleverly distribute tests across threads. For example, spread the fuzz tests on multiple threads, while running all unit tests single threaded. This also allowed for more precise types. Previously it looked like you could get a `DistributionReport` for unit tests, while in practice you can’t, for example.
- Running tests now collect`Debug.log` so that they can be displayed together with the test that made them. For unit tests, only the debug logs from the run that caused a fuzz test to fail is collected. (We ignore `Debug.log` from exploratory runs). Closes https://github.com/elm-explorations/test/pull/156. Closes https://github.com/elm-explorations/test/issues/154. Related: https://github.com/rtfeldman/node-test-runner/issues/400
- Running tests now report how long it took to execute.
- Make it possible for runners to run fuzz tests with the “fuzzer ints” (`RandomRun`) from a previous failure, allowing for an instant reproduction of a fuzz test failure. Related: https://github.com/rtfeldman/node-test-runner/issues/327
- There is now a function for identifying tests, making runners no longer need to do surgery on the compiled JS. Closes https://github.com/elm-explorations/test/pull/200


src/Test/Expectation.elm
```diff
@@ -1,55 +1,85 @@
...
-{-| Set the given (fuzz test input) of an expectation.
+{-| Due to backwards compatibility, we are forced to do this type conversion,
+without losing data – see `BreakingChangeWorkaround`.
 -}
-withGiven : String -> Expectation -> Expectation
-withGiven newGiven expectation =
-    case expectation of
-        Fail failure ->
-            Fail { failure | given = Just newGiven }
+fuzzExpectationToExpectation : FuzzTestExpectation -> Expectation
```
jfmengels: name suggestion: fromFuzzTestExpectation

src/Test/Expectation.elm
```diff
@@ -1,55 +1,85 @@
...
 
-{-| Create a failure without specifying the given.
--}
 fail : { description : String, reason : Reason } -> Expectation
 fail { description, reason } =
     Fail
-        { given = Nothing
-        , description = description
-        , reason = reason
+        { failData =
```
jfmengels: We can avoid creating a new record by not destructuring:

```elm
fail failData =
  Fail
    { failData = failData
    -- ...
    }
```

src/Test/Expectation.elm
```diff
@@ -1,55 +1,85 @@
...
+
+
+type alias FailData =
+    { description : String
+    , reason : Reason
+    }
+
+
+type FuzzTestExpectation
+    = FuzzTestPass { distributionReport : DistributionReport }
```
jfmengels: You could flatten this to `FuzzTestPass DistributionReport` (which I did in #258 if I'm not mistaken)

src/Test/Expectation.elm
```diff
@@ -1,55 +1,85 @@
...
         , reason : Reason
         , distributionReport : DistributionReport
+
+        -- This function runs the fuzzer and the fuzz test again,
+        -- with the input that caused this specific failure,
+        -- and throws away the result. The idea is that runners
+        -- can call this function to capture `Debug.log`s only
+        -- from the execution that caused the failure, drastically
+        -- reducing noise.
+        , rerunFailure : () -> ()
```
jfmengels, line 52: Could this function not return the result without throwing it away? The client of the API is here forced to call this function as if it's an explicit side-effect, which feels odd in Elm.

I imagine that the alternative is to return a value that is potentially a test pass, which is odd as well as the client would have to handle an impossible branch. Or could we potentially return a `FailData` (or whatever information we want to yield for a failure)?

src/Test/Fuzz.elm
```diff
@@ -59,23 +59,70 @@ validatedFuzzTest desc fuzzer getExpectation distribution =
...
 
-                Just failure ->
-                    { failure
-                        | expectation =
-                            failure.expectation
-                                |> Test.Expectation.withDistributionReport runResult.distributionReport
-                    }
-                        |> formatExpectation
+                    else
+                        -- Reproduce a previous failure.
```
jfmengels: Could you add a comment explaining how the previous failure is retrieved? It's not (yet) clear to me how that works. If that mechanism is described somewhere else, adding a comment pointing to that would be fine as well.

src/Test/Fuzz.elm
```diff
@@ -59,23 +59,70 @@ validatedFuzzTest desc fuzzer getExpectation distribution =
                         }
                         (initLoopState seed distribution)
             in
-            case runResult.failure of
-                Nothing ->
-                    Pass { distributionReport = runResult.distributionReport }
+            let
+                { failure, distributionReport } =
```
jfmengels, line 53: Nit: I could see this code block extracted to a function, could potentially clarify things.

src/Test/Internal.elm
```diff
@@ -84,3 +90,22 @@ duplicatedName tests =
...
     Elm.Kernel.Debug.toString
+
+
+runThunk : (a -> b) -> a -> Result String b
+runThunk =
+    Elm.Kernel.Test.runThunk
+
+
+wrapWithTryCatch : (a -> Expectation) -> (a -> Expectation)
+wrapWithTryCatch getExpectation =
```
jfmengels, line 101: Is `wrapWithTryCatch` mostly a rename of what was done before, in order to make it clearer what's happening?

src/Test/RunnerV2.elm
```diff
@@ -0,0 +1,498 @@
...
+
+
+{-| Run the unit test.
+-}
+runUnitTest : UnitTest -> UnitTestExpectation
+runUnitTest (UnitTest data) =
+    data.thunk ()
+
+
+{-| A unit test. It’s similar to a unit test, but has more data, and running
```
jfmengels:
```suggestion
{-| A fuzz test. It’s similar to a unit test but has more data, and running
```

src/Test/RunnerV2.elm
```diff
@@ -0,0 +1,498 @@
...
+it requires more arguments.
+
+`FuzzTest` is opaque for future extensibility. It contains the following:
+
+  - `tag : String`. The tag is set via [`tagTest`](#tagTest) and used by runners to cache test results.
+  - `labels : List String`. The list starts with the test name, and then contains each [`describe`](Test#describe) up the hierarchy.
+  - `runs : Maybe Int`. Contains the specified number of fuzz runs if [`fuzzWith`](Test#fuzzWith) was used.
+  - `thunk : Random.Seed -> Int -> List Int -> FuzzTestExpectation`. This function runs the fuzz test.
+
+Use the various `get*` functions to access each field, as well as [`runFuzzTest`](#runFuzzTest) to run it.
```
jfmengels:
```suggestion
Use the various `getFuzzTest*` functions to access each field, as well as [`runFuzzTest`](#runFuzzTest) to run it.
```

src/Test/RunnerV2.elm
```diff
@@ -0,0 +1,498 @@
...
+
+{-| A unit test.
+
+`UnitTest` is opaque for future extensibility. It contains the following:
+
+  - `tag : String`. The tag is set via [`tagTest`](#tagTest) and used by runners to cache test results.
+  - `labels : List String`. The list starts with the test name, and then contains each [`describe`](Test#describe) up the hierarchy.
+  - `thunk : () -> UnitTestExpectation`. This is the function originally passed to [`Test.test`](Test#test).
+
+Use the various `get*` functions to access each field, as well as [`runUnitTest`](#runUnitTest) to run it.
```
jfmengels:
```suggestion
Use the various `getUnitTest*` functions to access each field, as well as [`runUnitTest`](#runUnitTest) to run it.
```

src/Test/RunnerV2.elm
```diff
@@ -0,0 +1,498 @@
...
+
+{-| `FuzzTestFailData` is opaque for future extensibility. It contains the following:
+
+  - `distributionReport : DistributionReport`. See the [`DistributionReport`](Test.Distribution#DistributionReport) type.
+
+Use the various `get*` functions to access each field.
+
+-}
+type FuzzTestPassData
+    = FuzzTestPassData { distributionReport : DistributionReport }
```
jfmengels: Could be `FuzzTestPassData DistributionReport` to avoid creating an unnecessary record.

src/Test/RunnerV2.elm
```diff
@@ -0,0 +1,498 @@
...
+  - `reason : Reason`. See the [`Reason`](Test.Runner.Failure#Reason) type.
+  - `distributionReport : DistributionReport`. See the [`DistributionReport`](Test.Distribution#DistributionReport) type.
+  - `given : Maybe String`. This is the input to the test function that caused the failure, formatted with `Debug.toString`.
+    A fuzz test can in unusual circumstances fail to even produce a `given` value, which is why it is `Maybe`.
+  - `fuzzerInts : List Int`. This is the internal fuzzer state that produced `given`. A runner can pass this
+    to [runFuzzTest](#runFuzzTest) to reproduce a previous failure.
+  - `rerunFailure` is a function that runs the test function again with the input that
+    caused the failure. Runners can use this to capture `Debug.log` calls of the failing run.
+
+Use the various `get*` functions to access each field, as well as [`rerunFuzzTestFailure`](#rerunFuzzTestFailure) for running the failing run again.
```
jfmengels:
```suggestion
Use the various `getFuzzTestFail*` functions to access each field, as well as [`rerunFuzzTestFailure`](#rerunFuzzTestFailure) for running the failing run again.
```

src/Test/RunnerV2.elm
```diff
@@ -0,0 +1,498 @@
...
+                |> List.foldl
+                    (\subTest acc ->
+                        let
+                            sub =
+                                toTestsHelper tag labels subTest
+
+                            seenSkip =
+                                acc.seenSkip || sub.seenSkip
+                        in
+                        case ( acc.seenOnly, sub.seenOnly ) of
```
jfmengels: You could use nested if expressions to avoid allocating a tuple
```elm
if acc.seenOnly then
  if sub.seenOnly then
    -- ...
else
  if sub.seenOnly then
    -- ...
```

---

Alternatively, you could do checks like

```elm
if acc.seenOnly == sub.seenOnly then
    { unitTests = Array.append acc.unitTests sub.unitTests
    , fuzzTests = Array.append acc.fuzzTests sub.fuzzTests
    , seenSkip = seenSkip
    , seenOnly = acc.seenOnly
    }
else if acc.seenOnly then
  { acc | seenSkip = seenSkip } -- if necessary, see next comment
else
  { sub | seenSkip = seenSkip } -- if necessary, see next comment
```

src/Test/RunnerV2.elm
```diff
@@ -0,0 +1,498 @@
...
+                            ( False, False ) ->
+                                { unitTests = Array.append acc.unitTests sub.unitTests
+                                , fuzzTests = Array.append acc.fuzzTests sub.fuzzTests
+                                , seenSkip = seenSkip
+                                , seenOnly = False
+                                }
+
+                            ( True, False ) ->
+                                { acc
```
jfmengels: Could even avoid a record allocation by doing:
```elm
if acc.seenSkip == seenSkip then
  acc
else
  { acc | seenSkip = seenSkip }
```
(Though in the second branch a full record will be faster than a record update)


Same idea for `sub` in the `( False, True )` branch below

jfmengels (approved):
Looks very good to me, great work!

I left a bunch of improvement suggestions, but none are blocking.

src/Test/Internal.elm
lydell: Previously, the thunks weren’t “stored with a try-catch wrapper” (like I changed them to) – instead, when running a test thunk that happened in a try-catch. This doesn’t work with my new, more detailed fuzz test failures, though. Catching errors so late means that it’s much more trouble constructing a `FuzzTestFailData`. Whereas if we wrap early we get everything for free.

(Not sure if this explanation made sense at all.)

src/Test/Expectation.elm
lydell: It _could,_ yes. But what would it be used for? We already have all the data that it would return. And as you say, it allows for impossible cases:

- It could somehow return a pass.
- It could somehow fail to even generate a `given` value.

It does feel odd to have a side-effect function like this in Elm. But on the other hand, it is all about side effects: `Debug.log`s!

lydell: I found this PR: https://github.com/elm-explorations/test/pull/200

It made me realize that it’s possible to do stuff with Kernel code in this package, rather than doing ugly string replacements on the output in runners.

Maybe I can come up with a way that running a test includes the debug logs as data. Then it’s no need to expose this weird side-effect function.

lydell: I’ve changed the PR so that debug logs are collected by this package instead of by hacks in the runners. Things got much cleaner. (It mostly works, but is not quite done, but you can still take a peek if you’d like.)

PR #261, opened 2026-08-22
Isolate runtime exception tests

lydell:
The test uses `Debug.todo` which contains the line number of that `Debug.todo` in the error message. Having this test below a bunch of other tests is annoying, because if you make changes to one of the earlier tests, you might move the `Debug.todo` up or down a few lines and then have to update the expected error message.

This PR moves the runtime exception tests to their own file, decreasing the probability of having to update it due to unrelated changes.


jfmengels (approved):
Very nice change. I had to update this test quite a few times in my branch 😅

Janiczek (approved):

PR #272, opened 2026-09-22
RandomRun via TypedArrays

Janiczek:
## What 

Using a contiguous array for `RandomRun` instead of pure-Elm data structures.

The Elm wrapper is still pure - I'm using [`TypedArray.with`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/with) to return a new array instead of mutating the old one.

## Performance

`master`:
```
$ npm run test && hyperfine "node tests/elm.js"
...snip...
815.8 ms
```

This branch:
```
$ npm run test && hyperfine "node tests/elm.js"
...snip...
700.0 ms
```

## Compatibility

`Uint32Array` etc. are ES2015-era - okay.

[`TypedArray.prototype.with`](https://caniuse.com/mdn-javascript_builtins_typedarray_with) is the larger gamble (but still should be very reasonable, all of these are 3+ years old):

- [Node](https://nodejs.org/en/about/previous-releases) 20.0.0+
  - EOL since Mar 24, 2026
  - v22 is the oldest LTS version still actively supported
  - Node is now on v26
- browsers (if anybody wanted to make a HTML test runner):
  - Chrome/Edge 110+ (Feb 2023), Firefox 115+ (July 2023), Safari 16+ (Sep 2022)

Still, probably worth confirmation with @lydell re our Node promises to people.


lydell:
> Still, probably worth confirmation with @lydell re our Node promises to people.

node-test-runner currently supports Node.js 12. In that project, we only bump Node.js when there is a reason to do so, or when supporting an old version becomes too difficult. (It’s usually worth supporting a few EOL Node.js versions, since the Node.js that comes with some Linux distros lag behind.)

This sounds like a good reason to bump.

I also want somewhere around Node.js 20+ for a future node-test-runner change: I want to replace the `chokidar` dependency with plain `fs.watch`, and `fs.watch` only got good in more recent Node.js versions.

